PR #10, opened 2015-01-02
Updates to use not_thru flags

dnesbitt61:
Update costing to not allow using edges that are "not_thru" unless near the destination. 

@kevinkreiser @gknisely @dgearhart


kevinkreiser:
:ship:

PR #13, opened 2015-01-06
Add a CostFactory and Get auto routes to work

dnesbitt61:
Fixed a couple issues with PathAlgorithm (was calling wrong A* method!) and fix some pure virtual methods in Bicycle and Auto costing.

@dgearhart @gknisely @kevinkreiser

### DIFF
--- a/valhalla/thor/costfactory.h
+++ b/valhalla/thor/costfactory.h
@@ -1,0 +1,53 @@
+#ifndef VALHALLA_THOR_COSTFACTORY_H_
+#define VALHALLA_THOR_COSTFACTORY_H_
+
+#include <map>
+#include <memory>
+
+// Include derived cost methods?
+#include "thor/autocost.h"
+#include "thor/bicyclecost.h"
+#include "thor/pedestriancost.h"
+
+namespace valhalla {
+namespace thor {
+
+/**
+* Generic factory class for creating objects based on type name.
+*/
+template <typename T>
+class CostFactory {
+ public:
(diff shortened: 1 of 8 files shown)


gknisely:
ship it

valhalla/thor/costfactory.h
```diff
@@ -0,0 +1,53 @@
...
+#include "thor/bicyclecost.h"
+#include "thor/pedestriancost.h"
+
+namespace valhalla {
+namespace thor {
+
+/**
+* Generic factory class for creating objects based on type name.
+*/
+template <typename T>
```
kevinkreiser: whats the template for?

valhalla/thor/costfactory.h
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * Register
+   */
+  CostFactory& Register(const std::string& type, DynamicCost* obj) {
+    factory_funcs_.insert(std::make_pair(type, obj));
+    return (*this);
+  }
+
+  DynamicCost* Create(const std::string& type) const {
```
kevinkreiser: why must these be pointers? also if they must be pointers can we at least use shared_ptr so we dont have to worry about deallocating/copying them?

dnesbitt61: Yea, maybe you can help me with shared_ptr. I tried that (and the template was probably leftover from another Factory model I tried). I'm going to merge this and we can rework it later.

PR #19, opened 2015-01-07
added shape, shape offsets, and headings.

gknisely:
@dgearhart @dnesbitt61 @kevinkreiser

### DIFF
--- a/src/thor/trippathbuilder.cc
+++ b/src/thor/trippathbuilder.cc
@@ -105,7 +105,7 @@ TripPath TripPathBuilder::Build(GraphReader& graphreader,
 
     trip_edge->set_begin_heading(PointLL::HeadingAlongPolyline(points,30));
 
-    trip_edge->set_begin_heading(PointLL::HeadingAtEndOfPolyline(points,30));
+    trip_edge->set_end_heading(PointLL::HeadingAtEndOfPolyline(points,30));
     
   }
 
--- a/src/thor/trippathbuilder.cc
+++ b/src/thor/trippathbuilder.cc
@@ -93,7 +93,7 @@ TripPath TripPathBuilder::Build(GraphReader& graphreader,
     for (const auto& pt : points) {
       n = trip_shape.size();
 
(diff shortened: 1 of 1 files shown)


src/thor/trippathbuilder.cc
```diff
@@ -74,31 +81,42 @@ TripPath TripPathBuilder::Build(GraphReader& graphreader,
...
-    names = graphtile->GetNames(directededge->edgedataoffset(), names);
-    for (auto& name : names) {
-      std::cout << "   name=" << name << std::endl;
-    }
-    const std::shared_ptr<EdgeInfo> edgeinfo = graphtile->edgeinfo(
-        directededge->edgedataoffset());
-    // TODO - rm later
-    //edgeinfo->ToOstream();
-*/
+    trip_edge->set_begin_heading(PointLL::HeadingAtEndOfPolyline(points,30));
```
dgearhart: looks like copy/paste error - should be setting the end heading correct?

src/thor/trippathbuilder.cc
```diff
@@ -74,31 +81,42 @@ TripPath TripPathBuilder::Build(GraphReader& graphreader,
...
+      //Add shape in the reverse direction.
+      std::reverse(points.begin(), points.end());
+
+    }
+
+    trip_edge->set_begin_shape_index(trip_shape.size());
+
+    unsigned int n = 0;
 
+    for (const auto& pt : points) {
```
kevinkreiser: why use a foreach here when you need the index so bad? if there are dups we shoudl remove them at data creation time.

shouldnt we just do:
   if(is_reverse)
      trip_shape.assign(edgeinfo->shape().rbegin(), edgeinfo->shape().rend())
   else
      trip_shape.assign(edgeinfo->shape().begin(), edgeinfo->shape().end())

you are currently doing two copies here

kevinkreiser:
other than that shape thing :shipit:

dgearhart:
shape fix and then :+1:

dgearhart:
ship

PR #20, opened 2015-01-08
Updates becuase leg was removed from TripPath

dgearhart:
Also, passing TripPath into builder

@dnesbitt61 @kevinkreiser @gknisely


gknisely:
hate some of that formatting but u are the standards czar.  ship

PR #21, opened 2015-01-09
fix up factory for use in service

kevinkreiser:
this moves around a bunch of code but properly sets up the factory so that the caller doesnt have to know about the underlying implementation. the way it works is:

a user of the factory makes a factory object templated by the base class of the type of objects you want the factory to produce

you then register the various objects the factory can produce by sending the factory function pointers (callbacks) to methods that can create objects that derive from the templated base class above

the function pointers basically just point to the constructors of the pimpl for each heuristic

we should at some point add sending a boost ptree to these so as to be able to configure each ones individual options

### DIFF
--- a/src/thor/autocost.cc
+++ b/src/thor/autocost.cc
@@ -2,11 +2,82 @@
 
 #include <iostream>
 #include <valhalla/midgard/constants.h>
+#include <valhalla/baldr/directededge.h>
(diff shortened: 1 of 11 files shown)


dgearhart:
:+1:

dnesbitt61:
good stuff

PR #22, opened 2015-01-09
make trip builder method static since it has no internal state. also let...

kevinkreiser:
... it return the trip_path instead of reference because the compiler will likely copy elide it anyway


dnesbitt61:
ok

PR #25, opened 2015-01-12
changed to use DirectionsBuilder

dgearhart:
@dnesbitt61 @kevinkreiser @gknisely


dgearhart:
requires updates from odin

kevinkreiser:
please add tyr to your workspace. if my branch were merged there it would break from these changes too

dgearhart:
@kevinkreiser - I added tyr - thanks for the info

PR #27, opened 2015-01-13
Initial hierarchy transitioning

dnesbitt61:
Logic to transition up/down levels for auto routes and to limit expansion to improve performance. Distances and counts are currently hard-coded (for testing) - later will want config values and probably multi-pass relaxing of values if route fails.  Large performance improvement - still need to incorporate shortcut edges.


dgearhart:
:+1:

PR #30, opened 2015-01-16
Key fix to shortcut logic. Avoids superseded edges better.

dnesbitt61:
Improves performance by almost 50% on PA test route.


kevinkreiser:
:sheep: it

PR #32, opened 2015-01-16
update to use the correlated location in path finding

kevinkreiser:


kevinkreiser:
depends on baldr pr

gknisely:
:shipit:

dnesbitt61:
ok

PR #34, opened 2015-01-16
Connected Edges fixes and heading fix

dnesbitt61:
Some fixes to connected edge logic and how to determine opposing edgeto the incoming path edge at a node. Add a constant for the offset distance for finding heading (and use km as the method requires).


gknisely:
:shipit:

PR #36, opened 2015-01-19
Fix nodeinfo when transition up/down a level.

dnesbitt61:
Still need a way to find the opposing edge to the incoming edge when a level transition has occurred.


gknisely:
ship

dgearhart:
:+1:

PR #37, opened 2015-01-19
shape flipped.

gknisely:
fixed shape bug.  Was flipped.
@kevinkreiser @dgearhart @dnesbitt61


dgearhart:
:+1:  cool - now the turns should be correct

PR #41, opened 2015-01-21
Initial bicycle costing

dnesbitt61:
Needs a round of testing and also some config/options methods. Defer this until we start working on more extensive costing plugin development.

### DIFF
--- a/src/thor/bicyclecost.cc
+++ b/src/thor/bicyclecost.cc
@@ -79,7 +79,7 @@ class BicycleCost : public DynamicCost {
   virtual float UnitSize() const;
 
  protected:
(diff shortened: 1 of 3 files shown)


src/thor/bicyclecost.cc
```diff
@@ -70,14 +77,57 @@ class BicycleCost : public DynamicCost {
    * @return  Returns the unit size for sorting.
    */
   virtual float UnitSize() const;
+
+ protected:
+  enum BicycleType {
```
dgearhart: enum class BicycleType
to make scoped enum

dgearhart:
:+1:

PR #43, opened 2015-01-21
updates to use logging

kevinkreiser:


src/thor/adjlistbenchmark/adjlistbenchmark.cc
```diff
@@ -5,6 +5,7 @@
 #include <boost/program_options.hpp>
 
 #include <valhalla/midgard/util.h>
+#include <valhalla/midgard/logging.h>
```
kevinkreiser: updated this but its not been kept up with the internal apis so it doesn't build anyway

gknisely:
ship

kevinkreiser:
do the log levels look ok?

gknisely:
sure

src/thor/adjlistbenchmark/adjlistbenchmark.cc
```diff
@@ -127,25 +122,13 @@ int main(int argc, char *argv[])
   }
 
   if (vm.count("version")) {
-    std::cout << "valhalla " << VERSION << "\n";
+    std::cout << "AdjacencyListBenchmark " << VERSION << "\n";
```
dgearhart: still cout?

dgearhart:
otherwise :+1:

src/thor/adjlistbenchmark/adjlistbenchmark.cc
kevinkreiser: yep i purposely let all of the whole options and erroneous input to be cout and cerr because thats what normal linux programs usually look like. if you did:

ls some_non_existant_dir

you wouldnt want the output to be:

kkreiser@klaptop:~/sandbox/valhalla/tyr$ LANG=C ls some_non_existant_dir
2015/01/21 20:38:55.658508 [ERROR]  ls: cannot access some_non_existant_dir: No such file or directory

PR #47, opened 2015-01-23
Comment and log updates

dnesbitt61:
Update comments in AdjacencyList. Add pathcost and iteration count as Log info in FormPath. This is good info for making sure shortcuts and costs are making sense.


kevinkreiser:
:boat:

dgearhart:
:+1:

PR #50, opened 2015-01-26
Do not allow transition edges at origin. 

dnesbitt61:
May want to do this in Loki instead, but adding here for now.


gknisely:
:ship:

PR #51, opened 2015-01-26
Fix shortcuts!

dnesbitt61:
Only set has_shortcuts if one is actually allowed/taken. This makes a huge difference and fixes routes I was pondering for much too long.


gknisely:
:ship:

PR #59, opened 2015-02-03
Add initial code to form path on local level

dnesbitt61:
Update to use edgeinfo_offset for directed edges.
Update timing in pathtest to not do CPU clock usage but use std::chrono:high_resolution_clock so we get elapsed time (include I/O wait).

### DIFF
--- a/src/thor/formlocalpath.cc
+++ b/src/thor/formlocalpath.cc
@@ -1,0 +1,122 @@
+
+#include "thor/pathalgorithm.h"
+
+#include <valhalla/midgard/logging.h>
+
+using namespace valhalla::baldr;
+
+namespace valhalla {
+namespace thor {
+
+// TODO - this should be a config value (or get from config!)
+constexpr uint32_t kLocalLevel = 2;
+
+GraphId PathAlgorithm::GetStartNode(GraphReader& graphreader,
+                     const DirectedEdge* directededge) {
+  // Get the end node of directed edge
+  const GraphTile* tile = graphreader.GetGraphTile(directededge->endnode());
+  const NodeInfo* nodeinfo = tile->node(directededge->endnode());
+
+  // Get the opposing edge and then get its endnode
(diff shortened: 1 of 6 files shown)


src/thor/formlocalpath.cc
```diff
@@ -0,0 +1,122 @@
...
+        }
+if (!(directededge->endnode() == prior_local_node)) {
+  LOG_ERROR("Prior end not found...");
+}
+
+        // Set the prior end node on the local level to startnode
+        prior_local_node = startnode;
+
+        // Add the directed edge graph Id to the path
+        edgesonpath.emplace_back(GraphId(startnode.tileid(), startnode.level(),
```
dgearhart: can just use args

src/thor/pathtest/pathtest.cc
```diff
@@ -190,17 +203,25 @@ int main(int argc, char *argv[]) {
...
+  t2 = std::chrono::high_resolution_clock::now();
+  msecs = std::chrono::duration_cast<std::chrono::milliseconds>(t2-t1).count();
+  LOG_INFO("PathTest took " + std::to_string(msecs) + " ms");
 
   // Try the the directions
+  t1 = std::chrono::high_resolution_clock::now();
   TripDirections trip_directions = DirectionsTest(trip_path, originloc, destloc);
+  t2 = std::chrono::high_resolution_clock::now();
+  msecs = std::chrono::duration_cast<std::chrono::milliseconds>(t2-t1).count();
+  LOG_INFO("TripDirections took " + std::to_string(msecs) + " ms");
```
dgearhart: thanks for adding

dgearhart:
:+1:

src/thor/formlocalpath.cc
dnesbitt61: I'll get this on the next PR when I add the code to recover the shortcuts (just started that)

PR #61, opened 2015-02-04
TripPathBuilder Update for Exit Signs

dnesbitt61:
Change to access exits by directed edge index. Only store them if the directed edge exit flag is set.


dgearhart:
:+1:

PR #70, opened 2015-02-12
need a final node on this path

kevinkreiser:


dnesbitt61:
:ship: 
Even with 1.f rather than 1.0f

dgearhart:
:+1:

PR #71, opened 2015-02-13
partial distance and path finding edge cases

kevinkreiser:
this pr fixes a bunch of edge cases with partial distance and trivial pathes. if you start and end on the same link (and you are traveling in the right direction) the shape should be correct. it also fixes routes that end or start on nodes in terms of their shape (used to have duplicated points).

it doesnt fix loops, it doesnt fix pathfinding taking a longer route because it encounters the one destination edge first but has a worse partial distance than would have been if we waited to find the other edge with smaller partial distance. these will have to wait for another pr.

### DIFF
--- a/src/thor/pathalgorithm.cc
+++ b/src/thor/pathalgorithm.cc
@@ -7,6 +7,26 @@
 
 using namespace valhalla::baldr;
 
+namespace {
+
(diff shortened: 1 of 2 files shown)


dgearhart:
:+1:

dnesbitt61:
:shipit:

PR #73, opened 2015-02-17
Recover shortcuts on local path

dnesbitt61:
Working version (at least on a couple test routes). Still need to update pathalgorithm.cc to use it and figure out performance impact.

### DIFF
--- a/src/thor/formlocalpath.cc
+++ b/src/thor/formlocalpath.cc
@@ -164,8 +164,8 @@ GraphId PathAlgorithm::RecoverShortcut(GraphReader& graphreader,
   GraphId edgeid(startnode.tileid(), startnode.level(),
                  nodeinfo->edge_index());
   const DirectedEdge* directededge = tile->directededge(nodeinfo->edge_index());
-  for (uint32_t i = 0, n = nodeinfo->edge_count(); i < n;
-                  i++, directededge++, edgeid++) {
+  for (uint32_t i = 0; i < nodeinfo->edge_count(); i++,
+            directededge++, edgeid++) {
     if (directededge->trans_up() ||
         directededge->use() != shortcutedge->use() ||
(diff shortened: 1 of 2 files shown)


kevinkreiser:
looks good to me. is there any reason why this is in a seprate source file?

src/thor/formlocalpath.cc
```diff
@@ -118,5 +136,99 @@ if (!(directededge->endnode() == prior_local_node)) {
...
+  const NodeInfo* nodeinfo = tile->node(endnode);
+  const PointLL& endll = nodeinfo->latlng();
+
+  // Expand from the start node
+  tile = graphreader.GetGraphTile(startnode);
+  nodeinfo = tile->node(startnode);
+  GraphId edgeid(startnode.tileid(), startnode.level(),
+                 nodeinfo->edge_index());
+  const DirectedEdge* directededge = tile->directededge(nodeinfo->edge_index());
+  for (uint32_t i = 0, n = nodeinfo->edge_count(); i < n;
```
kevinkreiser: just a style thing but why instantiate n? just use nodeinfo->edge_count() directly, i'm certain its inlined. this way reads poorly (more like c?) in my opinion

dgearhart:
:+1:

src/thor/formlocalpath.cc
dnesbitt61: just a habit I got into...

dnesbitt61:
I put this in a separate file since I was not sure how large and complex it might be and whether we would actually keep it.

src/thor/formlocalpath.cc
kevinkreiser: no big deal just thought id mention it in case you were doing it subconciously

kevinkreiser:
cool makes sense

PR #74, opened 2015-02-18
allow for the loop route!

kevinkreiser:
there are cases where to get from a to b you basically have to go around the block and end on the same edge you begin on. the alogrithm doesnt let you touch the same edge twice so this pr works around that, and fixes a bittle part of the shape trimming stuff for partial distance.

oneway_loop


gknisely:
:ship:

dgearhart:
:+1:

dnesbitt61:
:shipit:

PR #75, opened 2015-02-18
take the shortest path even if its basically a crap shoot

kevinkreiser:
before these changes, this route was not the shortest route:

longer

with these changes we make sure to consider the shorter alternative as the proper shortest path:

shorter

it works by keeping track of what destination edges you've found and waiting until you've either, visited all edges in the set of possibly destination edges or (more likely) consider an edge which is has a larger cost to get to than the destination edge that has already been found.

### DIFF
--- a/src/thor/pathalgorithm.cc
+++ b/src/thor/pathalgorithm.cc
@@ -347,7 +347,6 @@ void PathAlgorithm::SetDestination(GraphReader& graphreader, const PathLocation&
 bool PathAlgorithm::IsComplete(const uint32_t edge_label_index) {
   //grab the label
   const EdgeLabel& edge_label = edgelabels_[edge_label_index];
-
   //if we've already found a destination and the search's current edge is more costly to get to, we are done
   if(best_destination_.first != kInvalidLabel && edge_label.truecost() > best_destination_.second)
     return true;
--- a/src/thor/pathalgorithm.cc
+++ b/src/thor/pathalgorithm.cc
@@ -57,7 +57,7 @@ PathAlgorithm::PathAlgorithm()
     : edgelabel_index_(0),
       adjacencylist_(nullptr),
       edgestatus_(nullptr),
-      best_destination_{kInvalidLabel, std::numeric_limits<float>::max(), 0}{
+      best_destination_{kInvalidLabel, std::numeric_limits<float>::max()}{
   edgelabels_.reserve(kInitialEdgeLabelCount);
 }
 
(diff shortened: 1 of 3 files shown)


src/thor/pathalgorithm.cc
```diff
@@ -328,21 +334,38 @@ void PathAlgorithm::SetOrigin(baldr::GraphReader& graphreader,
...
 // Add a destination edge
-void PathAlgorithm::SetDestination(const PathLocation& dest) {
-  // TODO - add partial distances
+void PathAlgorithm::SetDestination(GraphReader& graphreader, const PathLocation& dest, const std::shared_ptr<DynamicCost>& costing) {
+  // For each edge
   for (const auto& edge : dest.edges()) {
-    destinations_[edge.id] = edge.dist;
+    // Keep the id and the cost to traverse the partial distance
+    const GraphTile* tile = graphreader.GetGraphTile(edge.id);
+    destinations_[edge.id] = costing->Get(tile->directededge(edge.id)) * edge.dist;
```
kevinkreiser: we keep the id and also the partial cost of traversing this section of the edge

src/thor/pathalgorithm.cc
```diff
@@ -328,21 +334,38 @@ void PathAlgorithm::SetOrigin(baldr::GraphReader& graphreader,
...
-  // other direction
-  const auto& p = destinations_.find(edge_label.edgeid());
-  return p != destinations_.end();
+bool PathAlgorithm::IsComplete(const uint32_t edge_label_index) {
+  //grab the label
+  const EdgeLabel& edge_label = edgelabels_[edge_label_index];
+
+  //if we've already found a destination and the search's current edge is more costly to get to, we are done
+  if(best_destination_.first != kInvalidLabel && edge_label.truecost() > best_destination_.second)
+    return true;
```
kevinkreiser: we quit early if the next edge to be considered is already more expensive to reach than what we've found so far

src/thor/pathalgorithm.cc
```diff
@@ -328,21 +334,38 @@ void PathAlgorithm::SetOrigin(baldr::GraphReader& graphreader,
...
+  if(p != destinations_.end()) {
+    //if we didnt have another destination yet or this one is better
+    auto cost = edge_label.truecost() + p->second;
+    if(best_destination_.first == kInvalidLabel || cost < best_destination_.second){
+      best_destination_.first = edge_label_index;
+      best_destination_.second = cost;
+    }
+    destinations_.erase(p);
+    //if we've found all of the destinations we are done looking
+    return destinations_.size() == 0;
```
kevinkreiser: as we encounter destination edges we remove them, so if we haven't got any left we are done looking

dnesbitt61:
:ship:

PR #78, opened 2015-02-19
Set internal_intersection

dgearhart:
@dnesbitt61 @gknisely @kevinkreiser


dnesbitt61:
:ship:

gknisely:
After circle passes... :ship:

PR #79, opened 2015-02-20
added logic to check for access at nodes for each type of costing.

gknisely:
@dgearhart @kevinkreiser @dnesbitt61


dgearhart:
:+1:

PR #80, opened 2015-02-20
Bicycle costing changes

dnesbitt61:
Also sneak in some variable declaration changes that should make Kevin happy.


kevinkreiser:
:rocket:

PR #83, opened 2015-02-24
Fix exception when recovering shortcut edge on local level fails

dnesbitt61:
- when a shortcut edge was not recovered on the local level the tile was on the incorrect level. Use the start node (and get the tile of the start node) rather than saving nodeinfo and using the prior tile


dgearhart:
:+1:

PR #84, opened 2015-02-25
Add HierarchyLimits

dnesbitt61:
Structure that holds limits per level and does simple checks to see if expansion is allowed or transition is allowed. This allowed costing to be simplified. Next step is to hook in config settings to allow control of the limits and also allow multi-pass routes (e.g. if the first try of a route fails can increase limits).


gknisely:
:ship:

PR #85, opened 2015-02-26
Add property tree config to costing and use it for HierarchyLimits

dnesbitt61:
There are default hierarchy limits and now they can be overriden with property tree config. The plumbing to allow other costing options to be set/changed is also in place, though none are being used.

### DIFF
--- a/valhalla/thor/hierarchylimits.h
+++ b/valhalla/thor/hierarchylimits.h
@@ -1,6 +1,32 @@
 #ifndef VALHALLA_THOR_HIERARCHYLIMITS_H_
 #define VALHALLA_THOR_HIERARCHYLIMITS_H_
 
+// Default hierarchy transitions. Note that this corresponds to a 3 level
+// strategy: highway, arterial, local. Any changes to this will require
+// updates to the defaults.
+namespace {
+constexpr float kMaxDistance = std::numeric_limits<float>::max();
+
+// Default default maximum upward transitions (per level)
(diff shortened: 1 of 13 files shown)


dgearhart:
:+1:

valhalla/thor/hierarchylimits.h
```diff
@@ -26,13 +52,41 @@ struct HierarchyLimits {
...
+   */
+  HierarchyLimits(const boost::property_tree::ptree& pt, const uint32_t level)
+      : up_transition_count(0) {
+
+    // Construct string to identify the level of the hierarchy
+    std::string hl = "hierarchy_limits." + std::to_string(level);
+
+    // Set maximum number of upward transitions
+    max_up_transitions = pt.get<unsigned int>(hl + ".max_up_transitions",
+                    kDefaultMaxUpTransitions[level]);
```
kevinkreiser: could this ever be out of bounds or do we limit the hierarchy levels to a certain number somewhere?

kevinkreiser:
looks good other than comment

valhalla/thor/hierarchylimits.h
dnesbitt61: In dynamiccost.h it has 8 hardcoded with a TODO. Was going to use the max level constant (move it to graphcontants.h) but figured should find the highest level in the config and use that. So it will not overflow now

PR #88, opened 2015-03-02
Shortcuts - moved code out of autocost into pathalgorithm. 

dnesbitt61:
Need to evaluate if we need the code that prevents taking shortcuts near destination - but left in for now.


kevinkreiser:
:shipit:

PR #93, opened 2015-03-04
Use integer sizes for Adj list buckets

dnesbitt61:
Make sure mincost and current cost align to buckets.


kevinkreiser:
:camel: looks good

PR #94, opened 2015-03-05
Fix currentcost - should have been float! 

dnesbitt61:
Was leading to truncation errors in precision of costing.


kevinkreiser:
ship

dgearhart:
:+1:

PR #96, opened 2015-03-06
Costing Updates

dnesbitt61:
Update the costing model. Splits EdgeCost from TransitionCost (intersection/turn cost). Requires a Cost structure to be returned with both cost and seconds so we can handle time/schedules. Split out hierarchy transition edges so they either get skipped or added to the adjacency list (no need to go through other logic since they have no length).


dgearhart:
:shipit:

gknisely:
:christmas_tree:

PR #97, opened 2015-03-09
Use opp_local_idx for Uturn determination.

dnesbitt61:
Need to recreate data so the mjolnir fix that properly sets opp_local_idx on shortcut edges gets added.


dgearhart:
:shipit:

PR #98, opened 2015-03-09
Add method to relax hierarchy limits and allow multiple passes

dnesbitt61:
Also add a simple check to make sure route is converging and to return failure if not. Has to go 500K iterations without converging (pretty conservative!) but helps limit a long walking route that fails - so it fails sooner.

### DIFF
--- a/src/thor/pathtest/pathtest.cc
+++ b/src/thor/pathtest/pathtest.cc
@@ -41,6 +41,17 @@ TripPath PathTest(GraphReader& reader, const PathLocation& origin,
   t1 = std::chrono::high_resolution_clock::now();
   std::vector<GraphId> pathedges;
   pathedges = pathalgorithm.GetBestPath(origin, dest, reader, cost);
+  if (pathedges.size() == 0) {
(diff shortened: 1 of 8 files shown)


src/thor/pathtest/pathtest.cc
```diff
@@ -41,6 +41,17 @@ TripPath PathTest(GraphReader& reader, const PathLocation& origin,
   t1 = std::chrono::high_resolution_clock::now();
   std::vector<GraphId> pathedges;
   pathedges = pathalgorithm.GetBestPath(origin, dest, reader, cost);
+  if (pathedges.size() == 0) {
```
kevinkreiser: Is this going to happen automatically (future pr) and have the relaxtion factor come from config an array of factors for multiple passes or something?

dnesbitt61: Just starting to experiment with this. I think there will need to be some config control of # of passes and factor.

kevinkreiser:
:racehorse:

PR #99, opened 2015-03-10
Added AddTripIntersectingEdge to store the local intersecting edge information

dgearhart:
depends on odin

@dnesbitt61 @gknisely @kevinkreiser


gknisely:
:ship:

PR #100, opened 2015-03-11
Use shortcut mask rather than bool values

dnesbitt61:
so we only skip edges superseded by a matching shortcut edge.


dgearhart:
:shipit:

PR #102, opened 2015-03-31
Add inverse of bucketsize so we can avoid division

dnesbitt61:
Use static_cast in a couple places where old style cast was present

### DIFF
--- a/src/thor/adjacencylist.cc
+++ b/src/thor/adjacencylist.cc
@@ -3,16 +3,6 @@
 namespace valhalla {
 namespace thor {
 
-// Default constructor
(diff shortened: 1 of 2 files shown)


dgearhart:
:shipit:

src/thor/adjacencylist.cc
```diff
@@ -3,16 +3,6 @@
 namespace valhalla {
 namespace thor {
 
-// Default constructor
-AdjacencyList::AdjacencyList()
```
kevinkreiser: yay for cruft removal!

PR #104, opened 2015-04-02
Costing - auto and pedestrian updates for pt parsing

dnesbitt61:
Set default options, allow config JSON / property tree to override these. Added a method that can disable highway hierarchy level - need to test and see if htis is what we need to do in regions where highway hierarchy is unconnected/incomplete.


gknisely:
:ship:

PR #108, opened 2015-04-08
Use new SIF (costing) library.

dnesbitt61:
Moved costing into its own repository. This should allow easier additions by others.


kevinkreiser:
:boat:

PR #111, opened 2015-04-10
Change costing to costing_options

dnesbitt61:
Fix config (missing a ',')

### DIFF
--- a/src/thor/pathtest/pathtest.cc
+++ b/src/thor/pathtest/pathtest.cc
@@ -196,13 +196,13 @@ int main(int argc, char *argv[]) {
   }
 
   // Construct costing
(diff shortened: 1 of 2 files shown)


gknisely:
:ship:

src/thor/pathtest/pathtest.cc
```diff
@@ -214,7 +214,7 @@ int main(int argc, char *argv[]) {
 
   // Get the costing method - pass the JSON configuration
   std::shared_ptr<DynamicCost> cost = factory.Create(routetype,
-                       pt.get_child("costing." + routetype));
+                       pt.get_child("costing_options." + routetype));
```
dgearhart: Should we be using camel case for json?

PR #119, opened 2015-04-20
Updated route request files to use json 	

dgearhart:


kevinkreiser:
li.txt needs converted too right?

dgearhart:
good catch @kevinkreiser

kevinkreiser:
:rocket:

PR #126, opened 2015-04-28
Remove index from TransitionCost.

dnesbitt61:
Really wanted the localedgeidx of the directed edge. This fixes maneuver penalties on upper hierarchies.


dgearhart:
:shipit:

PR #132, opened 2015-05-08
Initial multimodeal path algorithm Accepts an array of costing methods.

dnesbitt61:
For now just switches between transit and pedestrian TravelModes. Also need to add transfer costs.
Depends on sif updates


dgearhart:
:shipit:

PR #133, opened 2015-05-11
first draft of thor service layer

kevinkreiser:
two things are added here, a binary that makes use of the thor service layer code, and the thor service layer code with is part of the thor library. it makes use of all the prime_server zmq deps but its part of the library so the simple server in tyr can just use it. this way the same code will be run regardless of in production or testing on our local machines

for thor it was really straight forward, grab some info out of the request that loki passes on to us. run the route, serialize the protobuf bytes (and the original request) onto odin.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ libtool
 # built objects
 pathtest
 citytest
+thor_service
 *.o
 .deps/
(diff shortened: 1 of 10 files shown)


dnesbitt61:
:shipit

PR #135, opened 2015-05-11
Updates to GetBestPathMM and pathtest

dnesbitt61:
Use costing=multimodal to populate array of costing methods per mode and use GetBestPathMM.


dgearhart:
:shipit:

PR #136, opened 2015-05-12
a few fixes for trippath builder and a log message

kevinkreiser:
most of this is whitespace but some of it isn't sorry for the hard to read diff. i basiclally reapplied diffs from 7 versions ago


dnesbitt61:
:ship:

dgearhart:
:shipit:

PR #139, opened 2015-05-12
Fix trivial routes ending at a node

dnesbitt61:
Destinations at a node need the incoming edge, not the outgoing. This fixes the destination edge list such that trivial routes ending at a node now work.


dgearhart:
:shipit:

dnesbitt61:
This also fixes routes that end at a dead end node!

gknisely:
:shipit:

PR #142, opened 2015-05-13
Added transit support, et al

dgearhart:


dgearhart:
depends on odin

dnesbitt61:
:shipit:

PR #147, opened 2015-05-15
Transit

gknisely:


src/thor/trippathbuilder.cc
```diff
@@ -635,7 +686,37 @@ TripPath_Edge* TripPathBuilder::AddTripEdge(const uint32_t idx,
...
+
+    TripPath_TransitInfo* transit_info = trip_edge->mutable_transit_info();
+    const TransitDeparture* transit_departure = graphtile->GetTransitDeparture(
+        idx, trip_id);
+    const TransitRoute* transit_route = graphtile->GetTransitRoute(
+        transit_departure->routeid());
+    const TransitTrip* transit_trip = graphtile->GetTransitTrip(trip_id);
+
+    //use route short name if available otherwise trip short name.
+    if (transit_route->short_name_offset())
```
dgearhart: Should we check if string names exist before we set in trip path

dgearhart:
:shipit:

PR #153, opened 2015-05-26
Multimodal updates

dnesbitt61:
Also add a method shared between the 2 BestPath algorithms to check if the new path to a temporarily labeled edge is lower cost than prior path.


gknisely:
:ship:

PR #155, opened 2015-05-26
throw exception if path_edges is 0 for multimodal; removed unneeded p…

kdiluca:
…ath algorithm clear since worker does cleanup
@kevinkreiser


dnesbitt61:
:shipit:

PR #156, opened 2015-05-27
Fix departure lookup for MulitModal

dnesbitt61:
Reuse of the tile parameter was an issue - if the end node crossed into a new tile it updated the tile and then tried to read departures from the wrong tile. Also the prior stop Id logic was changed to track the prior stop id until a new stop is entered.


gknisely:
:ship:

PR #160, opened 2015-05-29
Transit departure / arrival time updates

dnesbitt61:
Add block Id to trip edge. Update so that the arrive time is time arriving at current node and departure time is time departing from current node.


dgearhart:
:shipit:

PR #175, opened 2015-06-24
Add ModulateHierarchyLimits

dnesbitt61:
Modify arterial hierarchy distance from destination based on route length and (later) density at the destination.

### DIFF
--- a/valhalla/thor/pathalgorithm.h
+++ b/valhalla/thor/pathalgorithm.h
@@ -144,9 +144,9 @@ class PathAlgorithm {
   /**
    * Add edges at the origin to the adjacency list
    * @param  graphreader  Graph tile reader.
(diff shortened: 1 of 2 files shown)


valhalla/thor/pathalgorithm.h
```diff
@@ -137,16 +137,32 @@ class PathAlgorithm {
...
 
   /**
+   * Modify hierarchy limits.
+   */
+  void ModifyHierarchyLimits(const float dist, const uint32_t density);
+
+  /**
    * Add edges at the origin to the adjacency list
+   * @param  graphreader  Graph tile reader.
+   * @param  dest        Location information of the destination
```
gknisely: should be @param origin

gknisely:
:ship:

PR #177, opened 2015-06-26
Separate multimodal code

dnesbitt61:
Add method to test if destination cannot be reached. Does a shortest path test from the destination to see if a transit stop can be reached within the max mode distance at the destination (currently assumes pedestrian). This will prevent long search times on transit routes where there is no stop near the destination. Later when other travel modes are added the method will have to be extended.

Thor service and pathtest now do this check.


gknisely:
:ship:

PR #178, opened 2015-06-26
Assign motorway_junction and fork on node

dgearhart:


dgearhart:
depends on odin updates

PR #179, opened 2015-07-01
Cleanup and some simplification

dnesbitt61:
Add convenience method to add to adjacency list and temporarily label an edge. Remove edgelabel_index and just use size of the edgelabels vector. Remove graphreader from FormPath (not used).


gknisely:
:ship:

PR #184, opened 2015-07-08
Stastics for pathtest 

Kopkins:
Now when pathtest is run some statistics about the route are gathered. They are logged after the narrative and when pathtest is run using batch.sh the statistical output from each path is saved to statistics.csv

The various fields in the csv are as follows:
orgLat, orgLng, destLat, destLng, success, #Passes, runtime, trip time, length, arcDistance, #Manuevers

### DIFF
--- a/scripts/batch.sh
+++ b/scripts/batch.sh
@@ -57,8 +57,15 @@ sed -i -e "s/\([^\\]\)'|/\1|/g" -e "s/|'/|/g" "${TMP}"
 #from the log messages otherwise every line will be a diff
 #TODO: add leading zeros to output files so they sort nicely
 echo -e "\x1b[32;1mWriting routes from ${INPUT} with a concurrency of ${CONCURRENCY} into ${OUTDIR}\x1b[0m"
-cat "${TMP}" | parallel --progress -k -C '\|' -P "${CONCURRENCY}" "pathtest {} 2>&1 | grep -F NARRATIVE | sed -e 's/^[^\[]*\[NARRATIVE\] //' &> ${OUTDIR}/{#}.txt"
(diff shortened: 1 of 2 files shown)


src/thor/pathtest/pathtest.cc
```diff
@@ -382,6 +419,17 @@ int main(int argc, char *argv[]) {
     valhalla::midgard::logging::Configure(logging_config);
   }
 
+  // Something to hold all the stats
+  // INDEX  0       1       2       3       4       5       6       7         8       9           10
```
dgearhart: Define indexes as const so code is easier to follow i.e. ORG_LAT = 0, etc

scripts/batch.sh
```diff
@@ -57,9 +57,16 @@ sed -i -e "s/\([^\\]\)'|/\1|/g" -e "s/|'/|/g" "${TMP}"
...
 rm -f "${TMP}"
 
+for FN in $(ls ${OUTDIR}); do
+  cat ${OUTDIR}/${FN} | grep -F NARRATIVE | sed -e 's/^[^\[]*\[NARRATIVE\] //' > ${OUTDIR}/narr_${FN}
+  cat ${OUTDIR}/${FN} | grep -F STATISTICS | sed -e 's/^[^\[]*\[STATISTICS\] //' > ${OUTDIR}/stat_${FN}
+  cat ${OUTDIR}/${FN} | grep -F STATISTICS | sed -e 's/^[^\[]*\[STATISTICS\] //' >> ${OUTDIR}/STATISTICS.csv
+  rm ${OUTDIR}/${FN}
+done
+
 #if we need to run a diff
```
dgearhart: diff logic still works?

dgearhart:
:shipit:

PR #186, opened 2015-07-13
Fix destination partial distance 

dnesbitt61:
needs to be remainder of cost along destination edge and subtract this from the edge_label cost since cost
in the edge label is to the end of the destination edge.


gknisely:
:ship:

PR #187, opened 2015-07-14
Added program options for connectivity map and multi run

Kopkins:
Added two flags that can be passed to PathTest:
* `--connectivity` - Builds a connectivity map before running the path algorithms
* `--multi-run N` - Runs the path algorithm N additional times after the first run to demonstrate caching effects

__Note__ that the default behavior of the program has changed since these both ran on every execution prior to this update.

### DIFF
--- a/src/thor/pathtest/pathtest.cc
+++ b/src/thor/pathtest/pathtest.cc
@@ -76,7 +76,7 @@ namespace {
  */
 TripPath PathTest(GraphReader& reader, const PathLocation& origin,
                   const PathLocation& dest, std::shared_ptr<DynamicCost> cost,
-                  PathStatistics& data, bool multi_run) {
+                  PathStatistics& data, bool multi_run, uint32_t iterations) {
   auto t1 = std::chrono::high_resolution_clock::now();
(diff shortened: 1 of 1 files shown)


gknisely:
:ship:

kevinkreiser:
maybe `multi-run` should take an argument like the number of times to do so

dnesbitt61:
that sounds good. I had it set to 10 just to get a good average timing on a
route. But really didn't usually see much variability after the 1st run, so
usually sufficient to just run it a few times and average the elapsed time
values after the first.

On Tue, Jul 14, 2015 at 1:38 PM, Kevin Kreiser <notifications@github.com>
wrote:

> maybe multi-run should take an argument like the number of times to do so
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/valhalla/thor/pull/187#issuecomment-121318368>.
>



-- 
David Nesbitt
dave@mapzen.com
443-206-1819

Kopkins:
So should I add an argument to multi-run?

dnesbitt61:
yea, please

PR #188, opened 2015-07-14
Add an Update method for EdgeStatus.

dnesbitt61:
Calling Set again was corrupting the index. This didn't matter with forward A*, but need to retain the index in bidirectional A*.


gknisely:
:ship:

PR #189, opened 2015-07-16
Add bidirectional A*

dnesbitt61:
Derived bidirectional A* and multimodal from PathAlgorithm and create common interface for GetBestPath.
Use bidirectional A* for pedestrian routes > 10km.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -40,6 +40,7 @@ lib_LTLIBRARIES = libvalhalla_thor.la
 nobase_include_HEADERS = \
 	valhalla/thor/adjacencylist.h \
 	valhalla/thor/astarheuristic.h \
+	valhalla/thor//bidirectional_astar.h \
 	valhalla/thor/edgestatus.h \
 	valhalla/thor/pathalgorithm.h \
 	valhalla/thor/pathinfo.h \
@@ -48,8 +49,8 @@ nobase_include_HEADERS = \
 libvalhalla_thor_la_SOURCES = \
 	src/thor/adjacencylist.cc \
 	src/thor/astarheuristic.cc \
+	src/thor/bidirectional_astar.cc \
 	src/thor/edgestatus.cc \
-	src/thor/formlocalpath.cc \
 	src/thor/multimodal.cc \
 	src/thor/pathalgorithm.cc \
 	src/thor/trippathbuilder.cc \
--- a/Makefile.am
+++ b/Makefile.am
@@ -40,7 +40,7 @@ lib_LTLIBRARIES = libvalhalla_thor.la
 nobase_include_HEADERS = \
 	valhalla/thor/adjacencylist.h \
 	valhalla/thor/astarheuristic.h \
-	valhalla/thor//bidirectional_astar.h \
+	valhalla/thor/bidirectional_astar.h \
 	valhalla/thor/edgestatus.h \
 	valhalla/thor/pathalgorithm.h \
 	valhalla/thor/pathinfo.h \
--- a/src/thor/bidirectional_astar.cc
+++ b/src/thor/bidirectional_astar.cc
@@ -1,0 +1,377 @@
+#include <map>
+#include <algorithm>
+#include "thor/bidirectional_astar.h"
+#include <valhalla/baldr/datetime.h>
+#include <valhalla/midgard/logging.h>
+
+using namespace valhalla::baldr;
+using namespace valhalla::sif;
+
+namespace valhalla {
+namespace thor {
+
+// Default constructor
+BidirectionalAStar::BidirectionalAStar()
+    : PathAlgorithm() {
+}
+
+// Destructor
+BidirectionalAStar::~BidirectionalAStar() {
+  Clear();
+}
+
+// Clear the temporary information generated during path construction.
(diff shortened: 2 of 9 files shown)


Makefile.am
```diff
@@ -40,6 +40,7 @@ lib_LTLIBRARIES = libvalhalla_thor.la
 nobase_include_HEADERS = \
 	valhalla/thor/adjacencylist.h \
 	valhalla/thor/astarheuristic.h \
+	valhalla/thor//bidirectional_astar.h \
```
dgearhart: double slash

src/thor/bidirectional_astar.cc
```diff
@@ -0,0 +1,377 @@
...
+BidirectionalAStar::~BidirectionalAStar() {
+  Clear();
+}
+
+// Clear the temporary information generated during path construction.
+void BidirectionalAStar::Clear() {
+  edgelabels_.clear();
+  edgelabels2_.clear();
+  adjacencylist_.reset();
+  adjacencylist2_.reset();
```
gknisely: 2 means?

valhalla/thor/bidirectional_astar.h
```diff
@@ -0,0 +1,131 @@
...
+           const sif::TravelMode mode);
+
+  /**
+   * Clear the temporary information generated during path construction.
+   */
+  void Clear();
+
+ protected:
+  // A*, edge labels, adjacency list, and edge status for the reverse path
+  AStarHeuristic astarheuristic2_;
```
dgearhart: reverse instead of 2?

src/thor/bidirectional_astar.cc
```diff
@@ -0,0 +1,377 @@
...
+    }
+  }
+
+  // If we are here the route failed
+  return {};
+}
+
+// Convenience method to add an edge to the reverse adjacency list and
+// temporarily label it.
+void BidirectionalAStar::AddToAdjacencyList2(const GraphId& edgeid,
```
gknisely: 2, 2, 2, and 2

src/thor/service.cc
```diff
@@ -182,34 +184,51 @@ namespace {
...
-        return false;
+        // Construct just the requested costing model
+        if (costing == "auto" || costing == "bus") {
+          mode = valhalla::sif::TravelMode::kDrive;
+        } else if (costing == "bicycle") {
+          mode = valhalla::sif::TravelMode::kBicycle;
+        } else {
+          mode = valhalla::sif::TravelMode::kPedestrian;
+        }
+        mode_costing[static_cast<uint32_t>(mode)] = get_costing(request, costing);
```
kevinkreiser: shouldnt this be either `int` or `size_t`

src/thor/service.cc
```diff
@@ -182,34 +184,51 @@ namespace {
...
         mode_costing[2] = get_costing(request, "bicycle");
         mode_costing[3] = get_costing(request, "transit");
-        return true;
+        mode = valhalla::sif::TravelMode::kPedestrian;
       } else {
-        cost = get_costing(request, costing);
-        return false;
+        // Construct just the requested costing model
+        if (costing == "auto" || costing == "bus") {
+          mode = valhalla::sif::TravelMode::kDrive;
```
kevinkreiser: maybe it would be more practical to add the travelmode to the cossting class, so that costing controls it. in this way if we add new costing we get it automatically. otherwise we have to hardcode these in here for every new one there is.

valhalla/thor/bidirectional_astar.h
```diff
@@ -0,0 +1,131 @@
...
+   */
+  virtual ~BidirectionalAStar();
+
+  /**
+   * Form path between and origin and destination location using
+   * the supplied mode and costing method.
+   * @param  origin  Origin location
+   * @param  dest    Destination location
+   * @param  graphreader  Graph reader for accessing routing graph.
+   * @param  costing  An array of costing methods, one per TravelMode.
```
gknisely: should be  mode_costing

src/thor/pathtest/pathtest.cc
```diff
@@ -482,79 +482,86 @@ int main(int argc, char *argv[]) {
...
-    PathAlgorithm pathalgorithm;
-
-    // Have costing models use the per mode max distances
-    for (auto& cost : mode_costing) {
-      cost->UseMaxModeDistance();
+                        pt.get_child("costing_options.transit"));
+    mode = TravelMode::kPedestrian;
+    pathalgorithm = &mm;
+  } else {
+    if (routetype == "auto" || routetype == "bus") {
```
dgearhart: future path algorithm factory?

src/thor/service.cc
dnesbitt61: We keep mode separate to handle the multimodal case (which uses combinations of the costing classes).  Though we could keep it in each class but still need to pass it into GetBestPath

src/thor/service.cc
dnesbitt61: mode has to be TravelMode since it is passed into GetBestPath

dgearhart:
:shipit:

gknisely:
:ship:

PR #194, opened 2015-07-17
fixed bug where endnode was in another tile

gknisely:
fixed bug where endnode was in another tile and we were trying to get…the admin.  Error returned was GraphTile NodeInfo index out of bounds: 724724,2,28393 nodecount= 10307.


dnesbitt61:
:shipit

PR #196, opened 2015-07-20
Pathtest statistics handle failures differently now

Kopkins:
Failures caused by disconnectivity in the connectivity map are reported as `no_connectivity` and failures resulting from no route found are still reported as `false`.


dnesbitt61:
:shipit:

PR #197, opened 2015-07-20
Created two short scripts

Kopkins:
* run_city_routes.sh runs all all routes in city_to_city
* failCount.sh calculates a failure rate for output directories in pwd

### DIFF
--- a/scripts/batch.sh
+++ b/scripts/batch.sh
@@ -61,7 +61,7 @@ cat "${TMP}" | parallel --progress -k -C '\|' -P "${CONCURRENCY}" "pathtest {} 2
 rm -f "${TMP}"
 cat ${OUTDIR}/statistics.tmp | grep -F STATISTICS | sed -e 's/^[^\[]*\[STATISTICS\] //' &> ${OUTDIR}/statistics.csv
 rm ${OUTDIR}/statistics.tmp
-echo ${OUTDIR} >> outdir.txt
+echo ${OUTDIR} > outdir.txt
 
 #if we need to run a diff
 if [ -d "${DIFF}" ]; then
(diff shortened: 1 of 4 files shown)


scripts/run_city_routes.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+files=$(ls -b city_to_city)
+
+for f in ${files[@]}
+do
+  PATH=../:${PATH} ./batch.sh city_to_city/$f
```
dgearhart: Probably need batch.sh to output the OUTDIR to file so you can keep track of all the directories for a set of city to city tests and then append the OUTDIR to another file in this script so it can be used by the failCount.sh script

dgearhart: may want to output file name so we can see progress

scripts/failCount.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+directories=$(ls)
```
dgearhart: directories from file of OUTDIR created in run_city_routes.sh

scripts/failCount.sh
```diff
@@ -0,0 +1,15 @@
...
+for d in ${directories[@]}
+do
+  if [ -e $d/statistics.csv ]
+  then
+    cd $d
+    fails=$(cat statistics.csv | grep false | wc -l)
+    count=$(cat statistics.csv | wc -l)
+    perc=$(python -c "print ($fails/($count * 1.0))*100")
+    echo -n "Percent of failed routes = "
```
dgearhart: echo the directory so we know which percentage this corresponds to

scripts/batch.sh
```diff
@@ -61,6 +61,7 @@ cat "${TMP}" | parallel --progress -k -C '\|' -P "${CONCURRENCY}" "pathtest {} 2
 rm -f "${TMP}"
 cat ${OUTDIR}/statistics.tmp | grep -F STATISTICS | sed -e 's/^[^\[]*\[STATISTICS\] //' &> ${OUTDIR}/statistics.csv
 rm ${OUTDIR}/statistics.tmp
+echo ${OUTDIR} >> outdir.txt
```
dgearhart: i think this should just be echo ${OUTDIR} > outdir.txt (not concat)

dgearhart:
other than my last comment :shipit:

scripts/batch.sh
Kopkins: Definitely

dgearhart:
:+1:

PR #201, opened 2015-07-23
pathtest update when failure occurs

dnesbitt61:
If route fails list any origin and/or destination edges that are marked as unconnected.


dgearhart:
:shipit:

PR #204, opened 2015-07-27
Added failure cases to pathtest to provide better detail

Kopkins:
Possible status cases are:
* __success__ - occurs when route is successful
* __fail_no_route__ - occurs when no route can be found
* __fail_no_connectivity__ - occurs when there is no tile connectivity between origin and destination _only_ if the `--connectivity` flag is passed to pathtest
* __fail_invalid_origin__ - occurs when Loki cannot process origin 
* __fail_invalid_dest__ - occurs when Loki cannot process destination

The statistics.csv file generated by batch.sh now contains a header to help identify values.


dgearhart:
:ship:

dnesbitt61:
:shipit:

PR #207, opened 2015-07-29
Fix turn restriction bug

dnesbitt61:
Set restrictions on the origin edges so turn restrictions are honored off the initial edge.


gknisely:
:bow:

PR #208, opened 2015-07-30
Use bidirectional A* for bicycle over 10km.

dnesbitt61:
Fix destination in bidirectional A* (needed to add opposing edge to the adjacency list). Update to TransitionCostReverse call. 

Depends on sif changes

### DIFF
--- a/src/thor/service.cc
+++ b/src/thor/service.cc
@@ -82,8 +82,8 @@ namespace {
           thor::PathAlgorithm* path_algorithm;
           if (costing == "multimodal") {
             path_algorithm = &multi_modal_astar;
-          } else if (costing == "pedestrian") {
-            // Use bidirectional A* if over 10km
+          } else if (costing == "pedestrian" || costing == "bicycle") {
+            // Use bidirectional A* for pedestrian and bicycle if over 10km
             float dist = origin.latlng_.Distance(destination.latlng_);
             path_algorithm = (dist > 10000.0f) ? &bidir_astar : &astar;
           } else {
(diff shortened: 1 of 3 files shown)


dgearhart:
:shipit:

gknisely:
:bike:

src/thor/service.cc
```diff
@@ -82,8 +82,8 @@ namespace {
           thor::PathAlgorithm* path_algorithm;
           if (costing == "multimodal") {
             path_algorithm = &multi_modal_astar;
-          } else if (costing == "pedestrian") {
-            // Use bidirectional A* if over 10km
+          } else if (costing == "pedestrian" || costing == "bicycle") {
+            // Use bidirectional A* for pedestrian and bicycle if over 10km
             float dist = origin.latlng_.Distance(destination.latlng_);
             path_algorithm = (dist > 10000.0f) ? &bidir_astar : &astar;
```
kevinkreiser: is 10k the point at which performance is normally better with bidirectional? less than that unidirectional has less setup and bookkeeping and so is more efficient? just wondering if its empirically derived

kevinkreiser:
:see_no_evil:

src/thor/service.cc
dnesbitt61: It has more to do with making sure the special cases that the single tree A* handles (loop edge) are taken care of. We could test bidirectional for all these edge cases that can occur on short routes and make sure it works - then we could remove the distance check.

kevinkreiser: ah ok cool. this seems quite reasonable in the interim

PR #210, opened 2015-08-10
Alternate search based on cost

dnesbitt61:
In sparse regions the search in one direction can expand much more quickly than the other. Using cost helps level the search trees. Repairs issue sent in by Jess Crocker in remote mountain trail area.


gknisely:
:+1:

PR #212, opened 2015-08-11
use gcc/++4.9

kevinkreiser:
the point of this isnt for bleeding edge-ness but rather because regex (which became fully implemented in stl shipped with 4.9) is very useful and so other projects (particularly odin) will probably want to be making use of it.


dgearhart:
:shipit: when CI good

PR #226, opened 2015-09-04
FIx destination handling and some A* heuristics

dnesbitt61:
These commits fix issues submitted by Frane Saric, issue #220.

When a destination edge is added to the adjacency list its partial cost is added and the A* heuristic is set to 0 (used to be the full cost and heuristic from the end node of the edge). This means the first edge settled completes the route, removing the need for the more complex IsComplete method and tracking the "best destination". Code is simpler. 

The downside is that now every edge considered must be tested to see if it is a destination edge. This caused about a 10% performance drop. But changing to use a map rather than unordered_map (on the premise that the number of destination edges is usually very small) and using uint64_t rather than GraphId as the key in the EdgeStatus unordered_map ended up recovering the performance drop, so we are good there.


dgearhart:
:shipit:

kevinkreiser:
nice!

PR #232, opened 2015-09-15
Added side of street assignment to the trip path.

dgearhart:


dgearhart:
depends on odin

dnesbitt61:
:shipit

gknisely:
:star:

PR #233, opened 2015-09-16
Remove error log message

dnesbitt61:
Commented on how this can happen: (destination has been settled but we do not remove from dest_edges map). Validated some cases that it doesn't skip in error (e.g. lower cost found).


kevinkreiser:
:rocket:

PR #235, opened 2015-09-28
Remove special case loop and trivial logic. 

dnesbitt61:
Now the origin edges are not marked as settled/done so a path can come back onto them (around the block
case). This fixes some trivial route cases (begin and end on same edge) that were actually shorter than taking other roads. Removed the special case update_destinations method as this is now done in loki. But SetOrigin and SetDestination have to skip inbound/outbound edges if location is at a node.


kevinkreiser:
:ship: after we merge: https://github.com/valhalla/loki/pull/65

PR #238, opened 2015-09-30
Customize cost threshold in TimeDistanceMatrix constructor

ptpt:
So that you can use the cost threshold to control the search range.

### DIFF
--- a/valhalla/thor/timedistancematrix.h
+++ b/valhalla/thor/timedistancematrix.h
@@ -76,6 +76,17 @@ struct Destination {
 class TimeDistanceMatrix : public PathAlgorithm {
  public:
   /**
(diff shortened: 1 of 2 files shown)


valhalla/thor/timedistancematrix.h
```diff
@@ -76,6 +76,17 @@ struct Destination {
 class TimeDistanceMatrix : public PathAlgorithm {
  public:
   /**
+   * Constructor with default cost threshold.
+   */
+  TimeDistanceMatrix();
```
kevinkreiser: instead of having two constructors can we just have one that defaults the threshhold?:

`TimeDistanceMatrix(float initial_cost_threshhold = DEFAULT_COST_THRESHOLD);`

PR #240, opened 2015-10-02
Replace list with deque. Slight performance improvement.

dnesbitt61:
Update Bucket method to handle 3 cases (< currentbucket, within bucket range, or overflowbucket) and use in Add method and also to simplify the DecreaseCost method.


kevinkreiser:
:rocket:

PR #241, opened 2015-10-02
Allocate size for EdgeStatus

dnesbitt61:
Reserve size for the unordered_map (performance improvement on long routes).. Update EdgeStatus to use the strongly typed enum. Add convenience method to get the heuristic and distance in one call.


kevinkreiser:
:lollipop:

PR #242, opened 2015-10-02
Check if end node is in same tile - if so no need to get a new tile.

dnesbitt61:
Also pass distance to destination to HandleTransitionEdge (since we already have it and it is accessed several times in the method).


kevinkreiser:
:ship: it

Issue #245, opened 2015-10-07
Weird looking route (detour)


fnicollet:
Hello,

As I was trying out the API with the sample page, I tried this one:  36.442 , -119.95 , 35.778 , -119.467
Which gave me:
![image](https://cloud.githubusercontent.com/assets/1240481/10343832/38019c24-6d20-11e5-8090-b3d0ee76c3df.png)

The detour in the bottom-right corner looks strange, might be interesting to investigate

Fabien

dnesbitt61:
Will be part of our next release to the Mapzen turn by turn service.

PR #249, opened 2015-10-08
Skip long shortcut edges based on distance from the destination.

dnesbitt61:
Can lead to missing the proper connection if the shortcut is too long. Fixes issue #245


gknisely:
:+1:

PR #250, opened 2015-10-12
Update to allow more than 2 locations

dnesbitt61:


src/thor/service.cc
```diff
@@ -138,6 +167,64 @@ namespace {
...
+      // Find the path.
+      path_edges = path_algorithm->GetBestPath(origin, destination, reader,
+                                               mode_costing, mode);
+
+      // If path is not found try again with relaxed limits (if allowed)
+      if (path_edges.size() == 0) {
+        valhalla::sif::cost_ptr_t cost = mode_costing[static_cast<uint32_t>(mode)];
+        if (cost->AllowMultiPass()) {
+          // 2nd pass
+          LOG_INFO("Try again with relaxed hierarchy limits");
```
kevinkreiser: i'm not sure we should have this in the service since its generally headless. if its worth warning about to track as a possible issue maybe we should elevate it to log warn?

src/thor/trippathbuilder.cc
```diff
@@ -223,6 +223,33 @@ TripPath TripPathBuilder::Build(GraphReader& graphreader,
   if (origin.date_time_)
     tp_orig->set_date_time(*origin.date_time_);
 
+  // Add list of through locations
+  for (auto through : through_loc) {
+    TripPath_Location* tp_through = trip_path.add_location();
+    TripPath_LatLng* through_ll = tp_through->mutable_ll();
+    through_ll->set_lat(through.latlng_.lat());
+    through_ll->set_lng(through.latlng_.lng());
+    tp_through->set_type(
```
dgearhart: no need to check type - they are through locations - correct?

PR #253, opened 2015-10-19
Save the nodeinfo so we don't have to get it from the tile. 

dnesbitt61:
Use else if for the node type settings.


kevinkreiser:
:rocket: ha first for once!

PR #254, opened 2015-10-20
Fix error when destination is at a dead-end node.

dnesbitt61:
Was adding incorrect edge to the adjacency list - leading to a case where reverse search is immediately exhausted and forward search can never connect.
Also added a protection in case either direction search is exhausted.


dgearhart:
:shipit:

PR #256, opened 2015-10-20
Fix elapsed time along reverse path

dnesbitt61:
Fix issues with elapsed time on reverse path - one issue was floating point roundoff being accumulated and a bigger issue was incorrectly indexing edges in reverse path (off by one getting secs along an edge). 

Also, if either search is exhausted immediately return no path.


dgearhart:
:shipit:

PR #257, opened 2015-10-23
added units logic to return either mi or km for matrix

kdiluca:
Also changed input_locations to just locations so that it's consistent with routing.  Will update docs as well.


kevinkreiser:
merge #258 and then :ship:

PR #258, opened 2015-10-23
look into serialization bug, fix a few small things with unit parsing

kevinkreiser:
small stuff otherwise

### DIFF
--- a/src/thor/service.cc
+++ b/src/thor/service.cc
@@ -349,10 +349,12 @@ namespace {
       // Parse out units; if none specified, use kilometers
       double distance_scale = kKmPerMeter;
       auto units = request.get<std::string>("units", "km");
-      if (units == "km")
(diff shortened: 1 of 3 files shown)


src/thor/service.cc
```diff
@@ -373,29 +346,26 @@ namespace {
 
     //TODO: Do we need to pass costing for multimodal?
     worker_t::result_t  get_matrix(const MATRIX_TYPE matrix_type, const boost::property_tree::ptree &request, http_request_t::info_t& request_info) {
+      // Parse out units; if none specified, use kilometers
```
kevinkreiser: using `.get` you can give it a default. also since we are documenting this service seprately we dont have to support two types of input for km and two types of input for mi, so i would strongly favor only allowing either `km` or `mi` and not allowing spelling it out as well. imho that was a mistake in the regular route service

PR #265, opened 2015-10-30
Matrix: add check to return null for no route data

kdiluca:


src/thor/service.cc
```diff
@@ -66,12 +66,22 @@ namespace {
       const size_t origin, const size_t destination, const size_t start, const size_t end, double distance_scale) {
     auto row = json::array({});
     for(size_t i = start; i < end; i++) {
+      //check to make sure that data exists; if not, return nullptr for distance & time in matrix result
```
kevinkreiser: in this case its a check to see if a route was found

kevinkreiser:
:rocket:

PR #266, opened 2015-11-02
Add candidate connections so the least cost path is properly found. 

dnesbitt61:
Fix adding destinations (last commit broke the oneway destination edge case). Support hierarchies so we can apply to auto routes.


dgearhart:
wow - lots of changes :ship: when tests pass

PR #276, opened 2015-12-09
added check to see if the date that was inputed is before the tile_creation date

gknisely:
added check to see if the date that was inputed is before the tile_creation date    if so, forced to use assumed transit schedule


dgearhart:
:shipit:

dgearhart:
:+1:

PR #280, opened 2015-12-10
Added script to create route diffs

dgearhart:


dnesbitt61:
:shipit:

scripts/diff_routes.sh
```diff
@@ -0,0 +1,40 @@
...
+  echo "${DIFF_DIR} already exists"
+  exit 1
+fi
+
+CONCURRENCY=$(nproc)
+echo -e "\x1b[32;1mDiffing the output of ${OLD_DIR} with ${NEW_DIR} to ${OLD_DIR}_${NEW_DIR}_diff\x1b[0m"
+mkdir -p "${DIFF_DIR}"
+find ${OLD_DIR}/*.txt -printf "%f\n" | parallel --progress -P "${CONCURRENCY}" "diff ${OLD_DIR}/{} ${NEW_DIR}/{} > ${DIFF_DIR}/{}"
+                                                                                                       
+exit
```
kevinkreiser: not needed

kevinkreiser:
does this make the diffing that is in `batch.sh` obsolete?

dgearhart:
@kevinkreiser they are both needed
batch for typical run and diff
this diff script is used on analysis for city 2 city routes where we do not diff everyone

scripts/diff_routes.sh
dgearhart: this is so we do not create a diff if one already exists so we do not wipe out any diff notes, etc

PR #285, opened 2015-12-22
date time fix

gknisely:
date time is now set on the origin in set_origin if we are starting at the current time.

### DIFF
--- a/src/thor/bidirectional_astar.cc
+++ b/src/thor/bidirectional_astar.cc
@@ -167,9 +167,6 @@ std::vector<PathInfo> BidirectionalAStar::GetBestPath(PathLocation& origin,
         continue;
       }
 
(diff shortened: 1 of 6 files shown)


kevinkreiser:
is there a reason to do that common code in `set_origin` in both of the different path finding algorithms instead of once in, for example, the service before it calls either of them? just wonder if we can remove duplication by doing it further up stream

kevinkreiser:
other than comment looks good to me!

gknisely:
I am fine with that, but I don't want to do that now.  I will do it when we set the default to current if no type is set by the user.  Sound good?  Also, I checked in an arrive by fix.  Please review.

kevinkreiser:
looks good to me

PR #286, opened 2015-12-22
Do not enter not_thru regions in reverse search. 

dnesbitt61:
Could not do this within costing since many_to-one would not enter any not_thru regions which a valid origin might reside. This change in conjunction with the sif costing changes effectively results in the same logic.


dgearhart:
:shipit:

dnesbitt61:
fixes issue #284

PR #287, opened 2015-12-22
Simplify setting the origin timezone within SetOrigin

dnesbitt61:
Reuse nodeinfo (end node of an origin directed edge) to set the origin timezone rather than a special case at the start to get an edge, then its end node.


dgearhart:
:shipit:

gknisely:
:ship:

PR #293, opened 2016-01-11
added additional catch for unkown 500 errors

kdiluca:


src/thor/service.cc
```diff
@@ -183,12 +183,19 @@ namespace {
...
+          result.messages.emplace_back(response.to_string());
+          valhalla::midgard::logging::Log("500::" + std::string(e.what()), " [ANALYTICS] ");
+          return result;
+        }
         catch(...) {
           worker_t::result_t result{false};
           http_response_t response(500, "Internal Server Error", "Failed to parse intermediate request format", headers_t{CORS});
           response.from_info(info);
           result.messages.emplace_back(response.to_string());
-          valhalla::midgard::logging::Log("500::" + response.to_string(), " [ANALYTICS] ");
```
kevinkreiser: are you sure you dont want to still log this? if it starts happening we'd at least be able to know it, just put a hardcoded message in there like 'non-std::exception'  or?

kevinkreiser:
:rocket:

PR #294, opened 2016-01-12
added call to TransferCostFactor

gknisely:


gknisely:
depends on sif

dgearhart:
:shipit:

PR #295, opened 2016-01-12
Separate turn cost in the reverse path. 

dnesbitt61:
Apply this to the elapsed time logic in FormPath so the turn cost applies to the correct node. Also look for either temporary or permanently labelled edges when forming candidate connections - this fixes some connections (especially where the paths are in not_thru regions). Add special case logic to FormPath if the forward path connects to a destination edge (can happen when both locations are within the same not-thru region).


kevinkreiser:
:rocket:

PR #297, opened 2016-01-12
Add opposing edge offset to HandleTransitionEdgeReverse

dnesbitt61:
Fixes an issue with turn costs after transition edges


dgearhart:
:shipit:

PR #299, opened 2016-01-15
Include search time calculation when checking if long_request

kdiluca:


kdiluca:
Going to log at every level instead. Closing for now.

PR #300, opened 2016-01-18
Update edge label to have transition cost

dnesbitt61:
in the constructor/emplace calls. Separate downward transition logic and disable for now - see if that is what we want for driving routes. Use sort cost vs. cost as the factor for which path (forward or reverse) to expand. This PR depends on sif PR.

### DIFF
--- a/src/thor/bidirectional_astar.cc
+++ b/src/thor/bidirectional_astar.cc
@@ -13,7 +13,7 @@ namespace {
 GraphId GetOpposingEdgeId(const DirectedEdge* edge, const GraphTile* tile) {
   GraphId endnode = edge->endnode();
   return { endnode.tileid(), endnode.level(),
-           tile->node(endnode)->edge_index() + edge->opp_index() };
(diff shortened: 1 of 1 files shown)


src/thor/bidirectional_astar.cc
```diff
@@ -13,7 +13,7 @@ namespace {
 GraphId GetOpposingEdgeId(const DirectedEdge* edge, const GraphTile* tile) {
   GraphId endnode = edge->endnode();
   return { endnode.tileid(), endnode.level(),
-           tile->node(endnode)->edge_index() + edge->opp_index() };
+           tile->node(endnode.id())->edge_index() + edge->opp_index() };
```
gknisely: Did you mean to change this to .id()  Just double checking.

gknisely:
:ship:

src/thor/bidirectional_astar.cc
dnesbitt61: yes

PR #307, opened 2016-01-29
Back to using distance to destination in the StopExpanding method

dnesbitt61:
This solves some issues that occurred in auto routes in less urban areas.


gknisely:
:ship:

PR #311, opened 2016-02-03
Matrix logging updates

kdiluca:
-Added a many-to-many long request threshold in the config since this differs greatly from other route requests or the other 2 types of matrix requests


dnesbitt61:
:shipit:

PR #312, opened 2016-02-05
Make sure count_ is set first.

dnesbitt61:
It was not set for the case where there are 4 locations and TourCost is called with an indeterminant count_ value.


gknisely:
:+1:

kdiluca:
That fixed it, thanks!

PR #322, opened 2016-02-19
distance and weight updates.

gknisely:
added GetModeWeight from costing.  
Added call to UseMaxMultiModalDistance.  
Added logic for GetMaxTransferDistanceMM.


dnesbitt61:
:shipit:

gknisely:
depends on sif

PR #325, opened 2016-02-23
Add IsBestConnection.

dnesbitt61:
Split turn cost from newcost in forward path to get more accurate cost at the possible connection. Remove unused ModifyHierarchyLimitsReverse method.


kevinkreiser:
:rocket:

PR #326, opened 2016-02-24
Optimized order

kdiluca:
Please review the changes for optimized routing.  Keep in mind that the actual action name may change.  Also, I'm working on thor service tests so I don't want to merge the Optimized Order PRs until that is complete.  
NOTE:  There is now a revised header file for Thor and a lot of refactoring.
Thanks!


dgearhart:
:shipit:

src/thor/optimized_path.cc
```diff
@@ -0,0 +1,69 @@
...
+      TimeDistanceMatrix tdmatrix;
+      std::vector<TimeDistance> td = tdmatrix.ManyToMany(correlated, reader, mode_costing, mode);
+      std::vector<float> time_costs;
+      for (size_t i = 0; i < td.size(); i++) {
+        time_costs.emplace_back(static_cast<float>(td[i].time));
+      }
+      for (size_t i = 0; i < correlated.size(); i++) {
+        LOG_DEBUG("BEFORE reorder of locations:: " + std::to_string(correlated[i].latlng_.lat()) + ", "+ std::to_string(correlated[i].latlng_.lng()));
+      }
+      Optimizer optimizer;
```
kevinkreiser: i dont know if this keeps any state, or if its expensive to construct/destruct, if so though we'd probably want this to be a member of thor_worker_t

src/thor/optimized_path.cc
```diff
@@ -0,0 +1,69 @@
...
+    }
+    return input_locs;
+  }
+
+}
+
+namespace valhalla {
+  namespace thor {
+
+    worker_t::result_t  thor_worker_t::optimized_path(const std::vector<PathLocation> correlated, const std::string &costing, const std::string &request_str) {
```
kevinkreiser: correleated should be pass by reference

src/thor/service.cc
```diff
@@ -201,20 +75,23 @@ namespace {
         }
 
         // Initialize request - get the PathALgorithm to use
-        std::string costing = init_request(request);
+        std::string costing = thor_worker_t::init_request(request);
```
kevinkreiser: you dont need to do this, actually i'm not sure how this compiles. if you wanted to you could do `this->init_request` which i like stylistically but we havent been doing anywhere else so, probably dont want to

src/thor/optimized_path.cc
dnesbitt61: I think it is good like this. It keeps a tour and a best tour - so constructing it each time might be best (no need to init or clear)

src/thor/service.cc
```diff
@@ -588,9 +189,17 @@ namespace {
       return factory.Create(costing, *config_costing);
     }
 
-    std::string init_request(const boost::property_tree::ptree& request) {
+    std::string thor_worker_t::init_request(const boost::property_tree::ptree& request) {
```
kevinkreiser: same from above

src/thor/service.cc
```diff
@@ -653,35 +262,16 @@ namespace {
...
-    void cleanup() {
+    void thor_worker_t::cleanup() {
       astar.Clear();
       bidir_astar.Clear();
       multi_modal_astar.Clear();
       locations.clear();
       correlated.clear();
       if(reader.OverCommitted())
-        reader.Clear();
+      reader.Clear();
```
kevinkreiser: probably just an accident here

src/thor/trip_path.cc
```diff
@@ -0,0 +1,293 @@
...
+  namespace thor {
+
+  worker_t::result_t thor_worker_t::trip_path(const std::string &costing, const std::string &request_str, const boost::optional<int> &date_time_type){
+    worker_t::result_t result{true};
+    //get time for start of request
+    auto s = std::chrono::system_clock::now();
+    // Forward the original request
+    result.messages.emplace_back(request_str);
+
+    if (date_time_type && *date_time_type == 2) {
```
kevinkreiser: i think indents are messed up in this function

src/thor/trip_path.cc
```diff
@@ -0,0 +1,293 @@
...
+    return input_locs;
+  }
+
+}
+
+namespace valhalla {
+  namespace thor {
+
+  worker_t::result_t thor_worker_t::trip_path(const std::string &costing, const std::string &request_str, const boost::optional<int> &date_time_type){
+    worker_t::result_t result{true};
```
kevinkreiser: move this into both getPathArriveBy and getPathDepartFrom, since you have the request string going to both you can move the messages.emplace_back to each of them as well, then you wont need to pass this to them

src/thor/optimized_path.cc
```diff
@@ -0,0 +1,78 @@
...
+      auto trippaths = path_depart_from(best_order, costing, date_time_type, request_str);
+      for (const auto &trippath: trippaths){
+        result.messages.emplace_back(trippath.SerializeAsString());
+      }
+      //get processing time for thor
+      auto e = std::chrono::system_clock::now();
+      std::chrono::duration<float, std::milli> elapsed_time = e - s;
+      //log request if greater than X (ms)
+      if ((elapsed_time.count() / correlated.size()) > long_request_manytomany) {
+       LOG_WARN("thor::route optimized_path elapsed time (ms)::"+ std::to_string(elapsed_time.count()));
```
kevinkreiser: indent is wrong here

src/thor/service.cc
```diff
@@ -201,20 +81,26 @@ namespace {
         }
 
         // Initialize request - get the PathALgorithm to use
-        std::string costing = init_request(request);
+        std::string costing = this->init_request(request);
```
kevinkreiser: no need for `this` here as its implied. basically we arent doing this stylistically in the code so we might not want to start

src/thor/service.cc
```diff
@@ -201,20 +81,26 @@ namespace {
...
-          auto matrix_iter = MATRIX.find(*matrix);
-          if (matrix_iter != MATRIX.cend()) {
-            return get_matrix(matrix_iter->second, costing, request, info);
-          }
-          else { //this will never happen since loki formats the request for matrix
-            throw std::runtime_error("Incorrect matrix_type provided:: " + *matrix + "  Accepted types are 'one_to_many', 'many_to_one' or 'many_to_many'.");
-          }
-        }
-        return get_trip_path(costing, request_str, date_time_type);
+           if (*matrix != "optimized_order_route") {
```
kevinkreiser: this set of `if`s looks exploitable. can we have the `MATRIX` include an optimized enum as well and then use a switch case, all the matrix ones could go the `matrix` and the optimized could go to `optimized_path`, then the default would throw an exception. basically here its techinically possible that matrix=blah would get into the first if block and then it wouldnt return a matrix but it also wouldnt hit any exception because you have the throw inside of an else.

src/thor/service.cc
```diff
@@ -201,20 +81,26 @@ namespace {
...
+             valhalla::midgard::logging::Log("matrix_type::" + *matrix, " [ANALYTICS] ");
+             auto matrix_iter = MATRIX.find(*matrix);
+             if (matrix_iter != MATRIX.cend()) {
+               return thor_worker_t::matrix(matrix_iter->second, costing, request, info);
```
kevinkreiser: dont need `thor_worker_t::`

src/thor/service.cc
```diff
@@ -201,20 +81,26 @@ namespace {
...
+             }
+           } else if (*matrix == "optimized_order_route")  {
+               valhalla::midgard::logging::Log("matrix_type::" + *matrix, " [ANALYTICS] ");
+               return thor_worker_t::optimized_path(correlated, costing, request_str);
```
kevinkreiser: same as above

src/thor/service.cc
```diff
@@ -201,20 +81,26 @@ namespace {
...
+           } else {
+            //this will never happen since loki formats the request for matrix
+            throw std::runtime_error("Incorrect type provided:: " + *matrix + "  Accepted types are 'one_to_many', 'many_to_one' or 'many_to_many' or 'optimized_order_route'.");
+           }
+        } else
+          return thor_worker_t::trip_path(costing, request_str, date_time_type);
```
kevinkreiser: same as above

src/thor/service.cc
```diff
@@ -588,9 +198,17 @@ namespace {
...
+    std::string thor_worker_t::init_request(const boost::property_tree::ptree& request) {
       auto id = request.get_optional<std::string>("id");
-
+      // Parse out units; if none specified, use kilometers
+      double distance_scale = kKmPerMeter;
+      auto units = request.get<std::string>("units", "km");
+      if (units == "mi")
+        distance_scale = kMilePerMeter;
+      else {
+        units = "km";
```
kevinkreiser: this line isnt needed since it happens on line 205 does this

kevinkreiser:
i feel like arrive_by and depart_from are 95% the same. i wonder if we could just pass a flag to implement the differing bits of each

kdiluca:
ok, I'll take another look at arrive_by and depart_from.

dnesbitt61:
I'd rather hold of on the arrive_by and depart from for now. I'm not sure how much that will change as we actually start to put in more of the arrive by logic - so I prefer them separate until more fully implemented. Then we can pull them together if it still makes sense.

kdiluca:
ok, makes sense.

src/thor/optimized_path.cc
```diff
@@ -0,0 +1,77 @@
...
+        time_costs.emplace_back(static_cast<float>(td[i].time));
+
+      for (size_t i = 0; i < correlated.size(); i++)
+        LOG_DEBUG("BEFORE reorder of locations:: " + std::to_string(correlated[i].latlng_.lat()) + ", "+ std::to_string(correlated[i].latlng_.lng()));
+
+      Optimizer optimizer;
+      //returns the optimal order of the path_locations
+      auto order = optimizer.Solve(correlated.size(), time_costs);
+      std::vector<PathLocation> best_order;
+      for (size_t i = 0; i< order.size(); i++) {
```
kevinkreiser: are the `LOG_`s in this loop needed anymore? also why not a `for(auto index : order)`?

src/thor/service.cc
```diff
@@ -203,20 +85,25 @@ namespace {
...
-          }
-          else { //this will never happen since loki formats the request for matrix
-            throw std::runtime_error("Incorrect matrix_type provided:: " + *matrix + "  Accepted types are 'one_to_many', 'many_to_one' or 'many_to_many'.");
-          }
-        }
-        return get_trip_path(costing, request_str, date_time_type);
+        auto matrix_type = request.get_optional<std::string>("matrix_type");
+
+        if (matrix_type) {
+          if (*matrix_type == "one_to_many" || *matrix_type == "many_to_one" || *matrix_type == "many_to_many") {
```
kevinkreiser: this line is equivalent to lines 93 and 94 maybe you can find first and then you wont have to do string comparison (map will do it behind the scenes)

test/service
```diff
@@ -0,0 +1,228 @@
+#! /bin/bash
```
kevinkreiser: accidental checkin

kevinkreiser:
other than comments looks good! :checkered_flag:

kdiluca:
Finished work to have optimize routing use many-to-many limits; action name change; added a few test cases.
Please review.

PR #332, opened 2016-03-03
Do Not Track - Conditional Logging

kdiluca:
locked prime_server version to 0.3.2 and added DNT check for long_requests

### DIFF
--- a/src/thor/service.cc
+++ b/src/thor/service.cc
@@ -227,7 +227,7 @@ namespace {
       }
     }
 
-    worker_t::result_t get_trip_path(const std::string &costing, const std::string &request_str, boost::optional<int> &date_time_type, bool header_dnt){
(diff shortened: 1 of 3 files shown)


src/thor/service.cc
```diff
@@ -227,7 +227,7 @@ namespace {
       }
     }
 
-    worker_t::result_t get_trip_path(const std::string &costing, const std::string &request_str, boost::optional<int> &date_time_type){
+    worker_t::result_t get_trip_path(const std::string &costing, const std::string &request_str, boost::optional<int> &date_time_type, http_request_t::info_t& request_info){
```
kevinkreiser: this can be `const`

kdiluca:
@kevinkreiser , look ok?

dnesbitt61:
:shipit:

PR #335, opened 2016-03-11
Update EdgeLabels to include an operator index.

dnesbitt61:
This is an index into a map of operator names. This allows us to compare an operator to the predecessor operator and to add a penalty if not the same.


gknisely:
:tram:

PR #339, opened 2016-03-15
Use correct cost in Adjacency List 

dnesbitt61:
Add a GetOperatorId method. Update in-station vs. stop-to-stop transfer logic and if in-station make sure a transfer time is added and the departure is still valid (not in the past after adding the transfer time). Wrong value was being added to Adjacency List! Other changes include the way transfer cost is set and an operator change penalty is added.


gknisely:
:ship:

PR #342, opened 2016-03-22
First draft of Thor path algorithm description.

dnesbitt61:
Still more details to fill in but this gives an overview.


dgearhart:
:ship:

PR #346, opened 2016-03-29
Fix exception in AddTripIntersectingEdge

dnesbitt61:
fixes valhalla/baldr#236

Cannot get directed edge using local edge index when not on the local level. Only get directed edge for intersecting edges when on the local level (use it to set walkability and cyclability of intersecting edges).


dgearhart:
:shipit:

dnesbitt61:
Fixes issue: baldr#236

PR #348, opened 2016-03-31
Add get_path_algorithm method.

dnesbitt61:
Use bidirectional A* unless origin and destination share an edge.


dgearhart:
no more distance restriction for pedestrian and bicycle - correct?

dnesbitt61:
Still a max distance limit. No minimum distance limit so now it only uses A* for bus and if the origin/destination case share an edge (so it can do the trivial path logic).

dgearhart:
:shipit:

PR #349, opened 2016-03-31
Optimized - Fixed my ordering bug; also added logging for local testing for now; …

kdiluca:
…will remove later


dnesbitt61:
:shipit:

PR #350, opened 2016-04-11
Use a lower factor for the expansion within distance.

dnesbitt61:
This helps lower the long time to fail for some bidirectional routes.


gknisely:
:ship:

PR #353, opened 2016-04-26
Fixes to SetDestination. 

dnesbitt61:
Make sure the reverse A* heuristic is used and that the proper end node is used for the heuristic. Also make sure the opposing edge is set in the EdgeLabel. This fixes the strange bike route (loop) that was reported on Twitter.


kevinkreiser:
:rocket:

PR #356, opened 2016-05-03
updates for trusty

kevinkreiser:


dgearhart:
:shipit: after tests are good

PR #357, opened 2016-05-12
Update termination criteria and properly handle not_thru edges.

dnesbitt61:
Extend the searches after an initial connection has been found to allow for connections along long edges. Probably should be based on max edge cost but that reduces performance significantly so use a threshold based on number of iterations. Also - add not_thru edges to EdgeLabels and mark them, but do not add to adjacency list (effectively prunes search but allows connections on them for case where origin and destination connect at a shared node with both entering not_thru edges).


dgearhart:
:shipit:

dgearhart:
Ran several route files - the differences are what we expected :shipit:

PR #361, opened 2016-05-19
Fixes for best connection

dnesbitt61:
Fix not_thru logic. Fix best_connection (use full transition cost with penalties rather than just seconds). Override the not-thru flag on origin/destination to get around the case where a 2 edge loop (e.g. cul-de-sac) has not_thru improperly set.


gknisely:
:christmas_tree:

PR #363, opened 2016-05-20
trival path never had date_time set.

gknisely:


gknisely:
Just a copy and paste from below.

dnesbitt61:
:ship:

PR #370, opened 2016-05-25
Matrix & Optimized now using CostMatrix

kdiluca:
This updates calls the new CostMatrix::SourceToTarget function for all matrix and optimized requests to make response times faster

### DIFF
--- a/src/thor/matrix.cc
+++ b/src/thor/matrix.cc
@@ -8,7 +8,7 @@ using namespace prime_server;
 #include <valhalla/sif/autocost.h>
 #include <valhalla/sif/bicyclecost.h>
 #include <valhalla/sif/pedestriancost.h>
-#include <valhalla/thor/timedistancematrix.h>
+#include <valhalla/thor/costmatrix.h>
 
 #include "thor/service.h"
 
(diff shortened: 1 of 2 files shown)


dnesbitt61:
:shipit:

src/thor/matrix.cc
```diff
@@ -153,23 +153,32 @@ namespace valhalla {
...
-         matrix_action_type = "many-to-one";
-         break;
-       case MATRIX_TYPE::MANY_TO_MANY:
-         json = serialize_many_to_many(request.get_optional<std::string>("id"), correlated, tdmatrix.ManyToMany(correlated, reader, mode_costing, mode), units, distance_scale);
-         matrix_action_type = "many-to-many";
-         break;
-      }
+        case MATRIX_TYPE::ONE_TO_MANY:
+          source.push_back(correlated.front());
+          target.insert(target.begin(), correlated.begin(), correlated.end());
```
kevinkreiser: you can just use assignment here instead of insertion, it migth read better

kevinkreiser: in fact since they are all doing this maybe it would be better to just assign them at the begining and then erase what you dont need in each case except the last

src/thor/optimized_path.cc
```diff
@@ -39,8 +39,12 @@ namespace valhalla {
       // Forward the original request
       result.messages.emplace_back(std::move(request_str));
 
-      TimeDistanceMatrix tdmatrix;
-      std::vector<TimeDistance> td = tdmatrix.ManyToMany(correlated, reader, mode_costing, mode);
+      CostMatrix costmatrix;
+      std::vector<baldr::PathLocation> source;
```
kevinkreiser: same here you dont neven need to make  a source and target vector, simply use correlated for both in the call below

PR #373, opened 2016-05-26
Fix crash where adj. list current bucket is invalid on a subsequent Remove call.

dnesbitt61:
Also set threshold for a source or target to 0 after calling UpdateStatus so no extra calls are made to expand search in cases where adj. list is exhausted.


gknisely:
:ship:

PR #374, opened 2016-05-26
Do not decrement remaining sources or targets if already at 0

dnesbitt61:
(since unsigned int). This leads to no way to terminate the method.


gknisely:
:+1:

PR #378, opened 2016-06-01
Extend default threshold for CostMatrix.

dnesbitt61:
Also protect against failure to break out of loop - set a max number of iterations and log an error if this is reached.

### DIFF
--- a/src/thor/costmatrix.cc
+++ b/src/thor/costmatrix.cc
@@ -9,6 +9,7 @@ using namespace valhalla::sif;
 namespace {
 
 constexpr int kExtendSearchThreshold = 250;
+constexpr uint32_t kMaxMatrixIterations = 2000000;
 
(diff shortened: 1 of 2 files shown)


src/thor/costmatrix.cc
```diff
@@ -137,6 +138,13 @@ std::vector<TimeDistance> CostMatrix::SourceToTarget(
       LOG_INFO("SourceToTarget iterations: n = " + std::to_string(n));
       break;
     }
+
+    // Protect against edge cases that may lead to never breaking out of
+    // this loop. This should never occur but lets make sure.
+    if (n >= kMaxMatrixIterations) {
+      LOG_ERROR("Break out of CostMatrix::SourceToTraget loop: Exceeded max iterations");
```
kevinkreiser: should we maybe throw here so the service and other things calling this can catch the error and report it back to the user?

PR #383, opened 2016-06-02
Updates to EdgeLabel

dnesbitt61:
Prior transit stop does not work as an index (as same index can be used in different tiles) so use a GraphId instead. Update logic in multi-modal to use the node's GraphId for the transit stop. Also change terminology to path_distance (vs. walking_distance) for uses outside of multimodal.

This relies on sif changes.


gknisely:
:+1:

PR #386, opened 2016-06-13
Do not reserve EdgeLabels in PathAlgorithm

dnesbitt61:
This constructor is used in TimeDistanceMatrix. Both PathAlgorithm and TimeDistanceMatrix have been largely replaced by BidirectionalAstar and CostMatrix. No sense allocating this memory (which exists in some persistent objects in thor service.


kevinkreiser:
oh interesting approach! so you can make the object but it will only reserve once you use it. and since reserve does nothing if you already have enough its as efficient as possible.

:rocket:

PR #387, opened 2016-06-13
conform to refactor of connectivity map and path location

kevinkreiser:
this goes along with valhalla/baldr#254

### DIFF
--- a/src/thor/trippathbuilder.cc
+++ b/src/thor/trippathbuilder.cc
@@ -330,6 +330,7 @@ TripPath TripPathBuilder::Build(GraphReader& graphreader,
       end_pct = e.dist;
       end_sos = e.sos;
       end_vrt = e.projected;
+      break;
     }
   }
   // Set the destination side of street, if one exists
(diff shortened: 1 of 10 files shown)


dnesbitt61:
I like the methods for end_node and begin_node. Getting the "vertex" is a little klunky, but OK.
:shipit:

kevinkreiser:
indeed it does seem clunky but i figure once loki starts passing back results where the projected point changes (depending on island or matching score) then we'll need to think about how that affects the heuristic and come up with a proper way to handle multiple possible starts/goals

src/thor/trippathbuilder.cc
```diff
@@ -306,38 +306,36 @@ TripPath TripPathBuilder::Build(GraphReader& graphreader,
...
     }
   }
   // Set the destination side of street, if one exists
   if (end_sos != PathLocation::SideOfStreet::NONE)
     tp_dest->set_side_of_street(GetTripPathSideOfStreet(end_sos));
 
-  // Special case - destination is at a node - end percent is 1
-  if (dest.IsNode()) {
-    end_pct = 1.0f;
-  }
```
kevinkreiser: this seemed redundant, anyone know why its in there?

PR #388, opened 2016-06-21
Remove distance dependence from hieararchy logic

dnesbitt61:
This causes poor performance in very dense locations (like Los Angeles). Limit expansion on local and arterial based on number of upward transitions only. This matches the logic in CostMatrix which improves the chances of matching time/distance between matrix and point to point routes.


dgearhart:
This causes invalid directions for ferry_route.txt#5 - it does not stay on lower hierarchy to properly callout turn - missing maneuver#7
``` diff
< 7: Turn right to stay on North Ferry Road/NY 114. | 0.2 mi
<    VERBAL_ALERT: Turn right to stay on North Ferry Road.
<    VERBAL_PRE: Turn right to stay on North Ferry Road, New York 1 14.
<    VERBAL_POST: Continue for 2 tenths of a mile.
```

dgearhart:
:shipit: with the mjolnir changes

PR #392, opened 2016-07-06
avoid memory leaks

kevinkreiser:
this used to leak if you called compute and then didnt delete the thing it gave back. now, no matter what happens it will not leak the computed `isotile`


dnesbitt61:
:shipit:

Issue #391, opened 2016-07-05
Add Time Zone to Each Location's Time


dnesbitt61:
ISO 8601
Incoming requests do not need a timezone, but the outgoing response should attach a time zone to each location since we know the timezone at each node.

PR #394, opened 2016-07-11
Dst fix

gknisely:
updated to call new seconds_to_date function...we need to know when we cross dst begin or end time to correctly update the time.


gknisely:
Depends on baldr

gknisely:
fixes #391

dnesbitt61:
:shipit:

PR #398, opened 2016-07-14
Check if a transit directed edge is allowed by the transit costing method

dnesbitt61:
If use_bus or use_rail is false this will now avoid any lines of that type.


dgearhart:
:ship:

PR #399, opened 2016-07-15
nullptr_t → std::nullptr_t

bentley:
nullptr_t is part of the std namespace. This patch fixes a build error on OpenBSD.


kevinkreiser:
@bentley thanks again! although i fear we have a lot of this all over the place?

PR #406, opened 2016-07-27
Change to use a grid size in meters. 

dnesbitt61:
Adjust the bounds to allow more grid tiles in longitude than latitude (based on meters per degree of longitude at the first origin location's latitude).


kevinkreiser:
:shipit:

PR #407, opened 2016-07-28
Use the tile Intersect method 

dnesbitt61:
This is so we don't miss any tiles along the shape (without this can get gaps and disjoint contours). Also check if opposing directed edge has been settled - if so can skip marking tiles along the edge. Seems to be about a 30% speedup.


kevinkreiser:
:rocket:

PR #409, opened 2016-08-01
Add multimodal isocrhone support.

dnesbitt61:
Create a couple of common methods to simplify code a bit.

### DIFF
--- a/src/thor/isochrone.cc
+++ b/src/thor/isochrone.cc
@@ -1,7 +1,6 @@
 #include <iostream> // TODO remove if not needed
 #include <map>
 #include <algorithm>
-#include <limits>
 #include "thor/isochrone.h"
 #include <valhalla/baldr/datetime.h>
 #include <valhalla/midgard/distanceapproximator.h>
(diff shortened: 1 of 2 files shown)


kevinkreiser:
i dont understand the benefit to `Initialize` vs having this code in the constructor. i think it makes the interface confusing by presenting an option you dont want a person to actually be able to call multiple times, or do you want that?

dnesbitt61:
Initialize is called every time Compute (or ComputeMultiModal) - it clears the prior temporary information (edge labels, edge status, etc.). We can discuss how to properly use this within the service. But don't want the constructor to reserve the temporary data - only when it is initially used should it reserve memory.

kevinkreiser:
you can still remove the initialize method and get that effect. in fact, a better way migth be to remove the clear and function as well, put this back into the constructor and use a shared_ptr in the service.

kevinkreiser:
im just for simpler less java-y interfaces :wink:

kevinkreiser:
:rocket:

PR #410, opened 2016-08-02
add isochrone response and compute them in minutes not seconds

kevinkreiser:
this also removes the one long request parameter for matrix


dnesbitt61:
:shipit:

PR #411, opened 2016-08-04
Fix an issue with through locations when the location is at a node.

dnesbitt61:
We cannot set the origin edge to the through edge in this case - we want to remove the through edge and its opposing edge and keep all other edges.


gknisely:
:ship:

kevinkreiser:
:rocket:

PR #414, opened 2016-08-09
Sources to Targets for Matrix and Optimized

kdiluca:
- since all matrix location requests are being converted to sources and targets parameters, this allows us to remove a few of the matrix functions within the matrix.cc
- now uses the correlated sources and correlated targets
- if users use the locations parameters as input, then we parse the sources and targets back into locations for the response
- pulled out the logic to set correlated locations into a separate function

examples:
1) http://localhost:8002/sources_to_targets?json={%22sources%22:[{%22lat%22:40.71955054111941,%22lon%22:-74.00016546947882,%22type%22:%22break%22}],%22targets%22:[{%22lat%22:40.71955054111941,%22lon%22:-74.00016546947882,%22type%22:%22break%22},{%22lat%22:40.737892702684064,%22lon%22:-73.98368597729132,%22type%22:%22through%22},{%22lat%22:40.748557450215635,%22lon%22:-73.977506167721,%22type%22:%22through%22},{%22lat%22:40.727486422997785,%22lon%22:-73.98574591381475,%22type%22:%22through%22},{%22lat%22:40.75661990450192,%22lon%22:-73.97407294018194,%22type%22:%22through%22}],%22costing%22:%22auto%22}

compared to
2)  http://valhalla.dev.mapzen.com/one_to_many?json={%22locations%22:[{%22lat%22:40.71955054111941,%22lon%22:-74.00016546947882,%22type%22:%22break%22},{%22lat%22:40.737892702684064,%22lon%22:-73.98368597729132,%22type%22:%22through%22},{%22lat%22:40.748557450215635,%22lon%22:-73.977506167721,%22type%22:%22through%22},{%22lat%22:40.727486422997785,%22lon%22:-73.98574591381475,%22type%22:%22through%22},{%22lat%22:40.75661990450192,%22lon%22:-73.97407294018194,%22type%22:%22through%22}],%22costing%22:%22auto%22}&api_key=valhalla-t_16n1c


kevinkreiser:
something interesting i noticed when looking at your sample requests there is the `type` on each `location`. bascially `type` has no meaning for matrices. i wonder if we need to update the docs to say that type is ignored so dont bother with it for matrix of requests?

src/thor/matrix.cc
```diff
@@ -185,7 +142,7 @@ namespace valhalla {
       auto e = std::chrono::system_clock::now();
       std::chrono::duration<float, std::milli> elapsed_time = e - s;
       //log request if greater than X (ms)
-      if (!request_info.do_not_track && (elapsed_time.count() / correlated.size()) > long_request) {
+      if (!request_info.do_not_track && ((elapsed_time.count() / correlated_s.size()) || elapsed_time.count() / correlated_t.size()) > long_request) {
```
kevinkreiser: i wonder if this should be:

```c++
if (!request_info.do_not_track && elapsed_time.count() / (correlated_s.size() * correlated_t.size()) > long_request) {
```

this would be measuring the number of "routes" the matrix has to run and then the metric in the config would be avg time to do one of the routes in the matrix. seems more intuitive to me

src/thor/service.cc
```diff
@@ -258,18 +279,35 @@ namespace valhalla {
...
+        for(const auto& t : *request_targets) {
+          try{ locations.push_back(baldr::Location::FromPtree(t.second)); }
+          catch (...) { throw std::runtime_error("Failed to parse target"); }
+        }
+        correlated = store_correlated_locations(request, locations);
+
+        correlated_s.insert(correlated_s.begin(), correlated.begin(), correlated.begin() + request_sources->size());
+        correlated_t.insert(correlated_t.begin(), correlated.begin() + request_sources->size(), correlated.end());
+      }//we need something
+      else
```
kevinkreiser: i wonder if we need to protect against this case? i mean loki can never let it get here if it doesnt either use locations or sources and targets

PR #421, opened 2016-08-19
Added GetOffsetForHeading method that will adjust heading offset based on road class and 'use'

dgearhart:
depends on valhalla/baldr#273
fixes valhalla/odin#352
fixes valhalla/odin#296

``` Diff
< BEFORE
< 1: Drive south on Penwethers Lane. | 0.1 mi
> AFTER
> 1: Drive south on Penwethers Lane. | 0.1 mi
> 2: Turn left to stay on Penwethers Lane. | 0.0 mi
```

### DIFF
--- a/src/thor/trippathbuilder.cc
+++ b/src/thor/trippathbuilder.cc
@@ -20,9 +20,6 @@ using namespace valhalla::odin;
 
 namespace {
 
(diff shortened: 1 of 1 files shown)


gknisely:
:ship:

PR #425, opened 2016-08-25
Fixes for trivial pedestrian routes

dnesbitt61:
Switch to bidirectional A* for all pedestrian routes.  Updates for trivial cases where the start_pct > end_pct (e.g. reversed travel direction). This is needed for trivial pedestrian routes. Needed to update the start and end percent as well as get the opposing directed edge and update the side of street.


dgearhart:
:shipit:

PR #427, opened 2016-08-31
JSONify Error Response

kdiluca:


src/thor/service.cc
```diff
@@ -48,6 +48,26 @@ namespace {
 namespace valhalla {
   namespace thor {
 
+    worker_t::result_t thor_worker_t::jsonify_error(uint64_t code, const std::string& status, const std::string& error, http_request_t::info_t& request_info) const {
```
kevinkreiser: id make this an anonymous namespace function rather than a member since it doesnt need access to anything from the class instance

kevinkreiser:
:ship:

PR #432, opened 2016-09-14
added call to IsExcluded(tile, nodeinfo))

gknisely:


gknisely:
depends on sif and baldr

PR #433, opened 2016-09-19
Add support for not_thru

dnesbitt61:
Set the not_thru_pruning flag and check that it is true before pruning. Only used for costmatrix and bidirectional.

### DIFF
--- a/src/thor/bidirectional_astar.cc
+++ b/src/thor/bidirectional_astar.cc
@@ -206,7 +206,7 @@ std::vector<PathInfo> BidirectionalAStar::GetBestPath(PathLocation& origin,
       edgestatus_forward_->Update(pred.edgeid(), EdgeSet::kPermanent);
 
       // Prune path if predecessor is not a through edge
-      if (pred.not_thru()) {
+      if (pred.not_thru() && pred.not_thru_pruning()) {
         continue;
       }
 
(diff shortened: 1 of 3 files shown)


src/thor/bidirectional_astar.cc
```diff
@@ -637,7 +641,7 @@ void BidirectionalAStar::SetDestination(GraphReader& graphreader,
     AddToReverseAdjacencyList(opp_edge_id, sortcost);
     edgelabels_reverse_.emplace_back(kInvalidLabel, opp_edge_id, edgeid,
              opp_dir_edge, cost, sortcost, dist, opp_dir_edge->restrictions(),
-             opp_dir_edge->opp_local_idx(), mode_, c);
+             opp_dir_edge->opp_local_idx(), mode_, c, false);
```
kevinkreiser: so `false` means we are pruning from the destination on the reverse search path. im struggling to make sense of that. is that ok because if the route is taking place completely inside of a not through region then the forward path will find the destination edge?

kevinkreiser:
looks good to me. just my feeble brain cant quite comprehend all the possibilities when it comes to bidirectional and in or out of not through regions

src/thor/bidirectional_astar.cc
dnesbitt61: This just says that we do not start out allowing pruning (on either forward or reverse) until an edge that is thru is encountered - then we enable pruning.

PR #434, opened 2016-09-21
Costing option function updates

kdiluca:
- Updates to get_costing to no longer override the costing options


kevinkreiser:
:rocket:

PR #436, opened 2016-09-23
updates to pass travel_type and set in TripPath

dgearhart:


dnesbitt61:
:shipit:
With caveat that we may pull the type out of PathInfo so it can support multiple mode types better

dgearhart:
depends on valhalla/sif#144

PR #438, opened 2016-09-28
Enable distancematrix for one to many

xlqian:
Hello Valhalla! Greetings from Navitia! 

As we discussed in this issue: https://github.com/valhalla/thor/issues/408, we have been working on the api `one_to_many` for a couple of days. 

## **Why we are re-working on this**
(you're surely aware of all problems of your own baby :laughing: , this is only for the sake of traceability): 

- `one_to_many`/`many_to_one`/`many_to_many` api's `max_locations` is limited to 50 because of the huge demand  in RAM
- Response time is not satisfying for our use case 

## **What we have done in this PR**

As suggested in the  https://github.com/valhalla/thor/issues/408, we try enabling the `timedistancematrix.cc`, and used it only for `one_to_many` api. We noticed that this class is kind of abandoned, some private members and methods are missing, interfaces of outer classes are changed etc..., we did what we can to "repair" this class and magically it seemed that it worked.

## **What we have got**

The test result shows, with `timedistancematrix`, valhalla responded 3 times faster, however we found result returned by `timedistancematrix` were tiny different from the result computed by `costmatrix`, in 99% of cases, it was like 1 second for time or 1 meter for distance. We guessed that it may be due to the rounding errors. In 1% of cases, the difference was much larger. The check showed the result returned by `timedistancematrix` was correct.

## **What we are going to do**
Is to profile `timedistancematrix` and optimise it as we can

## **What we want to hear from you**

Besides your advices on code, we'd like to have your advices on how we switch the implementation of `one_to_many`/`many_to_one`/`many_to_many` APIs(https://github.com/valhalla/thor/compare/master...xlqian:enable_distancematrix_for_one_to_many?expand=1#diff-5856bbf06ad9bac8e0872ebf32f229c2R128). By arguments in URL? By valhalla.json? Or both?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -53,6 +53,7 @@ nobase_include_HEADERS = \
 	valhalla/thor/trippathbuilder.h \
 	valhalla/thor/service.h \
 	valhalla/thor/trafficalgorithm.h
+	valhalla/thor/timedistancematrix.h
 libvalhalla_thor_la_SOURCES = \
 	src/thor/astar.cc \
 	src/thor/astarheuristic.cc \
@@ -69,7 +70,8 @@ libvalhalla_thor_la_SOURCES = \
 	src/thor/route_action.cc \
 	src/thor/trippathbuilder.cc \
 	src/thor/service.cc \
-	src/thor/trafficalgorithm.cc
+	src/thor/trafficalgorithm.cc \
+        src/thor/timedistancematrix.cc
 
 libvalhalla_thor_la_CPPFLAGS = $(DEPS_CFLAGS) $(VALHALLA_DEPS_CFLAGS) @BOOST_CPPFLAGS@
 libvalhalla_thor_la_LIBADD = $(DEPS_LIBS) $(VALHALLA_DEPS_LIBS) $(BOOST_PROGRAM_OPTIONS_LIB) $(BOOST_FILESYSTEM_LIB) $(BOOST_SYSTEM_LIB) $(BOOST_THREAD_LIB) @PROTOC_LIBS@
--- a/Makefile.am
+++ b/Makefile.am
@@ -71,7 +71,7 @@ libvalhalla_thor_la_SOURCES = \
 	src/thor/trippathbuilder.cc \
 	src/thor/service.cc \
 	src/thor/trafficalgorithm.cc \
-        src/thor/timedistancematrix.cc
+	src/thor/timedistancematrix.cc
 
 libvalhalla_thor_la_CPPFLAGS = $(DEPS_CFLAGS) $(VALHALLA_DEPS_CFLAGS) @BOOST_CPPFLAGS@
 libvalhalla_thor_la_LIBADD = $(DEPS_LIBS) $(VALHALLA_DEPS_LIBS) $(BOOST_PROGRAM_OPTIONS_LIB) $(BOOST_FILESYSTEM_LIB) $(BOOST_SYSTEM_LIB) $(BOOST_THREAD_LIB) @PROTOC_LIBS@
--- a/src/thor/matrix_action.cc
+++ b/src/thor/matrix_action.cc
@@ -11,6 +11,7 @@ using namespace prime_server;
 
 #include "thor/service.h"
 #include "thor/costmatrix.h"
+#include "thor/timedistancematrix.h"
 
 using namespace valhalla;
 using namespace valhalla::midgard;
@@ -124,10 +125,17 @@ namespace valhalla {
         distance_scale = kMilePerMeter;
 
       json::MapPtr json;
-      //do the real work
-      thor::CostMatrix costmatrix;
-      json = serialize(matrix_type, request.get_optional<std::string>("id"), correlated_s, correlated_t,
-        costmatrix.SourceToTarget(correlated_s, correlated_t, reader, mode_costing, mode), units, distance_scale);
+      if (action == thor_worker_t::ONE_TO_MANY ){
+          thor::TimeDistanceMatrix timedistance_matrix;
+          json = serialize(matrix_type, request.get_optional<std::string>("id"), correlated_s, correlated_t,
+        		  timedistance_matrix.OneToMany(0,
+        				  correlated_t, reader, mode_costing, mode), units, distance_scale);
+      }else {
+          //do the real work
+          thor::CostMatrix costmatrix;
+          json = serialize(matrix_type, request.get_optional<std::string>("id"), correlated_s, correlated_t,
+            costmatrix.SourceToTarget(correlated_s, correlated_t, reader, mode_costing, mode), units, distance_scale);
+      }
 
       //jsonp callback if need be
       std::ostringstream stream;
--- a/src/thor/matrix_action.cc
+++ b/src/thor/matrix_action.cc
@@ -126,15 +126,14 @@ namespace valhalla {
 
       json::MapPtr json;
       if (action == thor_worker_t::ONE_TO_MANY ){
-          thor::TimeDistanceMatrix timedistance_matrix;
-          json = serialize(matrix_type, request.get_optional<std::string>("id"), correlated_s, correlated_t,
-        		  timedistance_matrix.OneToMany(0,
-        				  correlated_t, reader, mode_costing, mode), units, distance_scale);
-      }else {
-          //do the real work
-          thor::CostMatrix costmatrix;
-          json = serialize(matrix_type, request.get_optional<std::string>("id"), correlated_s, correlated_t,
-            costmatrix.SourceToTarget(correlated_s, correlated_t, reader, mode_costing, mode), units, distance_scale);
+        thor::TimeDistanceMatrix timedistance_matrix;
+        json = serialize(matrix_type, request.get_optional<std::string>("id"), correlated_s, correlated_t,
+          timedistance_matrix.OneToMany(0, correlated_t, reader, mode_costing, mode), units, distance_scale);
+      } else {
+        //do the real work
+        thor::CostMatrix costmatrix;
+        json = serialize(matrix_type, request.get_optional<std::string>("id"), correlated_s, correlated_t,
+          costmatrix.SourceToTarget(correlated_s, correlated_t, reader, mode_costing, mode), units, distance_scale);
       }
 
       //jsonp callback if need be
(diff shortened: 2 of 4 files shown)


dnesbitt61:
We are at our annual company meeting and will cmoment more on this tomorrow. Due to the way we use highway hierarchies it may be best to keep using Cost matrix for driving, but I think using TimeDistanceMatrix makes sense for bicycling and pedestrian

xlqian:
Thank you for your response! And Enjoy your annual meeting!

dnesbitt61:
Thanks for your patience. 

Can you let me know the types of one to many requests you tested to see this improvement? I think that for auto routes where locations are spread out somewhat, but still within say a 50 mile radius) that the CostMatrix method is faster and more accurately matches time and distance from our regular route method. This is due to the use of the highway hierarchies. 

So maybe in matrix action you can split out the special case of pedestrian and bicycle (which do not use highway hierarchies) and have them use TimeDistanceMatrix, while all others (including one_to_many for all other costing methods) use CostMatrix.

Maybe this approach can avoid having to make any config changes to select CostMatrix vs. TimeDistanceMatrix, but that would be my second option for the selection. I don't think we want any URL changes - outside users would probably be confused by this and it would be hard to explain/document.

TeXitoi:
Another very big problem with CostMatrix is that it uses a very big amount of memory for big requests (I think that we were at about 10Go for 1x200). The memory is principally in `DoubleBucketQueue`.

Our request are quite specific: the destinations are most of the time homogeneously distributed in a circle around the starting point. That's because we select the stop points at a given distance (computed as if we do a Superman 30 min ride at the speed of the selected mode) around the starting point.

xlqian:
### "Can you let me know the types of one to many requests you tested to see this improvement?"

The table below showed a comparison of performance with different types of  `one_to_many`requests: 

Request | CostMatrix Response Time| Used Memory | TimeDuration Matrix Response Time | Used Memory 
-------- | ---|------------ | ---- | ---
Car with 500 destinations | Crashed(lack of memory) | >10GB | 6094ms | ~100MB
Car with 200 destinations    | 4682ms|~4GB | 1425ms | ~50MB
Pedestrian with 500 destinations     | 8164ms | ~8GB | 1815ms | ~50MB
Pedestrian with 200 destinations     | 5806ms | ~4GB | 1240ms | ~ 50MB

Here is the link to dowload these requests: https://www.wetransfer.com/downloads/b18e8ac1fe70e5e960d37dbc1c7cdd2a20161003104611/3faf623560767c81131d698d625b7ed820161003104612/5bc964

In most cases, the time and distances returned by two methods are quite similar.... 

**Note**: we also observed that the memory is constantly increasing every time a `one_to_many` request is received, which caused a crash after 3 requests of "Car with 200 destinations" for example...

dnesbitt61:
I am fine using the updated TimeDistanceMatrix for pedestrian and bicycle within one_to_many matrices. We will need to look at auto a bit more due to the hierarchy transitions. Another possibility is allowing an optional use of TimeDistanceMatrix for auto or when number of locations exceeds some threshold. There are other ways to reduce memory with CostMatrix such as using a larger bucket size to decrease the memory allocated for the double bucket queue or relying on the overflow bucket and initializing with a smaller cost range.

TeXitoi:
Why did you chose this `DoubleBucketQueue` implementation and not something more classic as a Fibonacci heap?

dnesbitt61:
Long ago I saw this recommended in a paper as the fastest way to implement the adjacency list in Dijkstra. I have worked with heap sorting before, but not Fibonacci heaps. Benchmarks and other performance analysis I've done through the years has always shown double bucket queue to be faster. We have a small benchmark in tools that compares STL priority_queue to the DoubleBucketQueue implementation. Note also that this does not test any methods that decrease cost of an existing edge, which I have heard is expensive in some heap implementations.

antoniomedrano:
Fibonacci heaps are well known to perform poorly on real-world data. While Fib-heaps have the best worst-case complexity on pathological networks, they also are quite slow on just about every other possible network type. It takes a very specially crafted network to make Fib-heaps faster than other priority queues. But on those special networks, other priority queues will exhibit extremely bad behavior. Fortunately, these special networks never occur in reality.

TeXitoi:
I don't understand: this PR is marked merged, but nothing of it is on master branch.

dnesbitt61:
With Valhalla 2.0 we are updating the graph tile structure and especially the hierarchy. We no longer duplicate edges across hierarchies, now motorways are stored only on highway hierarchy (level 0) and not duplicated on the local and arterial levels.

I think we merged your changes (e.g. master now has the DoubleBucketQueue and timedistancematrix is in Makefile.am) with the exception of actually enabling it in matrix_service. We fixed some issues with CostMatrix and still want to use it in production except possibly for pedestrian.

TeXitoi:
OK. Merging xlqian:enable_distancematrix_for_one_to_many just changed matrix_action.cc, so I think your right.

But I have strange results. I try now with 1.1.4 to see if that's a problem of timedistancematrix with the new tile format.

dnesbitt61:
Did you make new data? If so and are still seeing strange results I may have some time to test this more with the new tile structure.

TeXitoi:
Yes. And 1.1.4 segfault at data import.

dnesbitt61:
I tested tools valhalla_run_matrix to use TimeDistanceMatrix rather than CostMatrix and it worked (not all matched since I used auto with some moderate lengths. 

Several repos have been updated in master as we prepare a 2.0 release. For building data need to make sure baldr and mjolnir are both updated.

TeXitoi:
My problem seems on a pedestrian path (just a point to point path). I have lots of errors in the creations of the tiles of the kind
```
2016/11/04 15:30:48.578573 [ERROR] No opposing edge at LL=45.7581,3.06319 Length = 1 Startnode 0/3015/146470 EndNode 0/3015/146470 Shortcut 0
2016/11/04 15:30:48.578582 [WARN] No regular edges found from end node
2016/11/04 15:30:48.578666 [ERROR] No opposing edge at LL=45.7639,3.12751 Length = 1 Startnode 0/3015/146497 EndNode 0/3015/146497 Shortcut 0
2016/11/04 15:30:48.578677 [WARN] No regular edges found from end node
2016/11/04 15:30:48.579029 [ERROR] No opposing edge at LL=45.7518,3.10922 Length = 1 Startnode 0/3015/146648 EndNode 0/3015/146648 Shortcut 0
2016/11/04 15:30:48.579040 [WARN] No regular edges found from end node
2016/11/04 15:30:48.579276 [ERROR] No opposing edge at LL=45.7786,3.09454 Length = 1 Startnode 0/3015/146750 EndNode 0/3015/146750 Shortcut 0
2016/11/04 15:30:48.579287 [WARN] No regular edges found from end node
2016/11/04 15:30:48.579356 [ERROR] No opposing edge at LL=45.8063,3.12768 Length = 1 Startnode 0/3015/146773 EndNode 0/3015/146773 Shortcut 0
2016/11/04 15:30:48.579366 [WARN] No regular edges found from end node
2016/11/04 15:30:48.579458 [ERROR] No opposing edge at LL=45.8134,3.1213 Length = 1 Startnode 0/3015/146805 EndNode 0/3015/146805 Shortcut 0
2016/11/04 15:30:48.579468 [WARN] No regular edges found from end node
```

The trace of the server on the request:
```
$ ~/dev/valhalla-relwithdeb/local/bin/valhalla_route_service valhalla.json 1
2016/11/04 15:48:04.930233 [INFO] Tile extract successfully loaded
0 2016/11/04 15:48:23.123667 POST /route HTTP/1.1
2016/11/04 15:48:23.123847 [INFO] Got Loki Request 0
2016/11/04 15:48:23.124099 [ANALYTICS] location_count::2
2016/11/04 15:48:23.124115 [ANALYTICS] costing_type::pedestrian
2016/11/04 15:48:23.124230 [ANALYTICS] location_distance::1.764426km
2016/11/04 15:48:23.132770 [INFO] Got Thor Request 0
2016/11/04 15:48:23.133027 [ANALYTICS] travel_mode::1
2016/11/04 15:48:23.133043 [ANALYTICS] #_passes::1
2016/11/04 15:48:23.524030 [INFO] path_cost::67737
2016/11/04 15:48:23.524075 [INFO] FormPath path_iterations::5,331219
2016/11/04 15:48:23.524346 [ANALYTICS] admin_state_iso::
2016/11/04 15:48:23.524366 [ANALYTICS] admin_country_iso::
2016/11/04 15:48:23.524498 [WARN] thor::route trip_path elapsed time (ms)::391.461060
2016/11/04 15:48:23.524517 [WARN] thor::route trip_path exceeded threshold::{"costing_options":{"pedestrian":{"walking_speed":"4.032000000000001"},"bicycle":{"bicycle_type":"Hybrid"}},"locations":[{"lat":"45.787495","type":"break","lon":"3.097788"},{"lat":"45.798955","type":"break","lon":"3.113493"}    ],"costing":"pedestrian","directions_options":{"units":"kilometers"},"\/route":"","action":"0","correlated_0":{"edges":[{"id":"12763300650552","dist":"0.4291986","projected":{"lon":"3.097804","lat":"45.78749"},"sos":"0","score":"1.499737"},{"id":"2425045934344","dist":"0.5708014","projected":{"lon":"3.097804","lat":"45.78749"},"sos":"0","score":"1.499737"}        ],"location_index":"0"},"correlated_1":{"edges":[{"id":"19494017719864","dist":"0.5350732","projected":{"lon":"3.113521","lat":"45.79897"},"sos":"0","score":"3.020482"},{"id":"21573285207608","dist":"0.4649268","projected":{"lon":"3.113521","lat":"45.79897"},"sos":"0","score":"3.020482"}        ],"location_index":"1"}}

2016/11/04 15:48:23.524529 [ANALYTICS] valhalla_thor_long_request_route
2016/11/04 15:48:23.524772 [INFO] Got Odin Request 0
2016/11/04 15:48:23.525095 [INFO] trip_path_->node_size()=43
2016/11/04 15:48:23.550560 [INFO] maneuver_count::29
2016/11/04 15:48:23.550962 [INFO] Got Tyr Request 0
2016/11/04 15:48:23.551143 [ANALYTICS] language::en-US
2016/11/04 15:48:23.551227 [ANALYTICS] trip_time::68658s
2016/11/04 15:48:23.551235 [ANALYTICS] trip_length::76.799324km
0 2016/11/04 15:48:23.551986 200 12625
```

But that's may not be the good place to talk about that.

TeXitoi:
With another pbf that's working. Sorry for the noise.

PR #451, opened 2016-10-17
Go back to using A* if any origin and destination edges are the same

dnesbitt61:
This fixes issue #450  with through locations when there is only 1 edge in the path. Also, bus should use bidirectional A* as all others.


dgearhart:
:ship:

PR #453, opened 2016-10-19
Trace actions

kdiluca:
Refactored to create factory in trace_actions and fixed leak by deleting pointer

### DIFF
--- a/src/thor/trace_route_action.cc
+++ b/src/thor/trace_route_action.cc
@@ -6,7 +6,6 @@ using namespace prime_server;
 #include <valhalla/baldr/geojson.h>
 #include <valhalla/baldr/pathlocation.h>
 #include <valhalla/baldr/errorcode_util.h>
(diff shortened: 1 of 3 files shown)


src/thor/trace_route_action.cc
```diff
@@ -41,10 +41,11 @@ worker_t::result_t thor_worker_t::trace_route(
   result.messages.emplace_back(request_str);
 
   // Call Meili for map matching to get a collection of pathLocation Edges
+  meili::MapMatcherFactory factory(config);
```
kevinkreiser: i wonder if it would be worth it for us to write a clear function for this factory object. i mean i wonder if it holds the cache of the edges' shapes gridded up and if so how it would negatively impact performance. need to review more code in meili or check @ptpt 's work in the mapmatch service code to see how its doing it

dgearhart:
:ship:

PR #454, opened 2016-10-20
Branch 1.1.2

mtehver:
I am doing PRs for the changes I made to make Valhalla more mobile friendly. The changes are based on 1.1.2 release, so there could be some conflicts with the HEAD. Most changes are quite simple to help with discrepancies between C++ compiler/stdlib versions. I added support for compiling Valhalla with either boost::regex (the default) or std::regex (if USE_STD_REGEX preprocessor variable is set). I also abstracted graph storage handling into separate classes by refactoring filesystem based code from baldr GraphTile/GraphReader into abstract GraphTileStorage and GraphTileFsStorage. The storage handling code should perhaps include also write support (that would make mjolnir codebase cleaner), but that is not currently implemented.

### DIFF
--- a/src/thor/costmatrix.cc
+++ b/src/thor/costmatrix.cc
@@ -2,6 +2,7 @@
 #include <algorithm>
 #include "thor/costmatrix.h"
 #include <valhalla/midgard/logging.h>
+#include "config.h"
 
 using namespace valhalla::baldr;
 using namespace valhalla::sif;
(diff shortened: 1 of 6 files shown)


kevinkreiser:
see comments on baldr#308

kevinkreiser:
we're going to close this for now. when you have time feel free to open another pr. let us know if you have any questions

PR #459, opened 2016-11-02
Valhalla 2.0

kevinkreiser:
This changes all the things that are in GraphTiles. It's not that huge of a departure but we've taken care to reorganize all the bits so that growth will be possible without breaking compatibility. We've changed the notion of what is in the tiles in that now we dont duplicate the higher class edges across all tile levels. Only low class in level 2, only middle class in level 1, and only high class in level 0.


dnesbitt61:
:shipit:

PR #462, opened 2016-11-09
Fix the separation of transition costs in the reverse path. 

dnesbitt61:
Improperly added entire cost, rather than separating out the elapsed_time portion so the reverse path can be properly constructed with elapsed time to the end of each edge.


dgearhart:
:shipit:

PR #466, opened 2016-11-18
if a bus or rail line goes beyond the max, we need to consider transit options

gknisely:
fixed a bug where the isochrone was cut off if we took transit that was greater than the max_seconds.

before

image

after

image

issue reported by @meghanhade

### DIFF
--- a/src/thor/isochrone.cc
+++ b/src/thor/isochrone.cc
@@ -389,8 +389,7 @@ std::shared_ptr<const GriddedData<PointLL> > Isochrone::ComputeMultiModal(
   uint32_t max_transfer_distance = 99999.0f; //costing->GetMaxTransferDistanceMM();
 
   // Initialize and create the isotile
-  // add 5 minutes in case the transit trip is right on the edge.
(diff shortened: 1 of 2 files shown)


src/thor/isochrone.cc
```diff
@@ -389,7 +389,8 @@ std::shared_ptr<const GriddedData<PointLL> > Isochrone::ComputeMultiModal(
   uint32_t max_transfer_distance = 99999.0f; //costing->GetMaxTransferDistanceMM();
 
   // Initialize and create the isotile
-  auto max_seconds = max_minutes * 60;
+  // add 5 minutes in case the transit trip is right on the edge.
+  auto max_seconds = (max_minutes * 60) + 300;
```
dnesbitt61: Can you add 10 minutes to the input to this and to the regular Compute call? If you do this within isochrone_action.cc (add 10 to contours.back) it would be centralized and applied to both methods. Either that or also add the same to the regular Compute call.

kevinkreiser:
:rocket:

PR #475, opened 2016-11-29
Refactored RouteMatcher

dgearhart:


kevinkreiser:
:rocket: id prefer the 3 static methods to be in an anonymous namespace and hidden but its not a deal breaker to me

PR #479, opened 2016-11-29
 Add complex restriction check. 

dnesbitt61:
Fixes "trivial" path test where destination is on same edge as origin but in opposing direction of oneway travel. This case falls back to A* logic and needs to honor complex restrictions.


gknisely:
:ship:

PR #480, opened 2016-11-30
Move increment of index to before while loop 

dnesbitt61:
(index is used as a reference as the next correlated shape index).


dgearhart:
:ship:

PR #486, opened 2016-12-02
Cost along destination should use the forward directed edge 

dnesbitt61:
(not the opposing directed edge). Fixes high path_cost when forward and reverse speeds are set separately (and the opposing one is 0).

@smWork @gknisely


gknisely:
:ship:

PR #492, opened 2016-12-05
Fixes duplicate call to meili::MapMatcher::OfflineMatch

### LINKED ISSUE #489: Map matching should only be run once per trace sequence
Currently, meili::MapMatcher::OfflineMatch is run for every coordinate in a trace sequence (i.e. for a sequence of 100 coordinates, map matching is run 100 times).

ericrwolfe:
Fixes #489.


dnesbitt61:
Thanks so much!

PR #493, opened 2016-12-05
Pass custom map-matching parameters to meili

### LINKED ISSUE #490: Trace actions should support custom meili map-matching parameters
Meili allows customizing parameters for map-matching, but it's not currently possible to pass those through to meili via the trace route actions.

ericrwolfe:
Fixes #490.

Combined with the `meili.customizable` array in [conf/valhalla.json](https://github.com/valhalla/conf/blob/master/valhalla.json#L60), allows passing of custom map-matching parameters to meili. 

Adds an additional `trace_options` attribute to the json payload for `trace_route` and `trace_attributes` actions, and merges with the meili defaults defined in the conf file. Currently, only accepts float parameters.

For example:
```
POST http://localhost:3002/trace_route
{
	"costing":"auto",
	"encoded_polyline":"...",
	"trace_options":{
		"search_radius":50,
		"sigma_z":10,
		"gps_accuracy":10
	}
}
```


kevinkreiser:
@ericrwolfe can you reopen this and we can fix the issue that we ran into?

ericrwolfe:
@kevinkreiser sure. Can you be more specific on how you'd like to proceed? I can't reopen this or #490 on my end, but I can create a new PR or issue if you'd like.

kevinkreiser:
yeah sadly the mechanics for it arent really on github. although i think you can indeed get the patch for this and apply it via `git am patch_file`. you can then edit it to fix the issue (make trace_options optional) and open a new pr.

PR #498, opened 2016-12-06
Use actual distance along the shape

dnesbitt61:
Use actual distance along shape when rejecting edges (rather than distance from prior node to a shape point). This allows rejection of edges that might be a U-turn vs. the shape which goes around the block and never exceeds the threshold distance. For efficiency it computes lengths between shape points once and reuses those when computing lengths along portions of the shape.

Also removes the check for start and end being along the same edge.


dgearhart:
:ship:

kevinkreiser:
@ericrwolfe this may be a more robust way of solving #494

PR #499, opened 2016-12-06
Revert "Pass custom map-matching parameters to meili"

kevinkreiser:
Reverts valhalla/thor#493

@ericrwolfe the previous PR requires that the request have `trace_options` in it. we need to use get optional so that its not always required to be in the request


kevinkreiser:
we'll get the changes back in with a bit more testing

ericrwolfe:
@kevinkreiser ah, sorry about that.

PR #500, opened 2016-12-07
Pass custom map-matching parameters to meili

ericrwolfe:
Improves #493 (reverted by #499) by making `trace_options` optional.


src/thor/service.cc
```diff
@@ -244,6 +242,41 @@ namespace valhalla {
 
     }
 
+    void thor_worker_t::parse_trace_config(const boost::property_tree::ptree& request) {
+      auto costing = request.get<std::string>("costing");
+      trace_config.put<std::string>("mode", costing);
+
+      std::unordered_set<std::string> customizable;
```
kevinkreiser: instead of filling `customizable` out on every request can we follow the model in the worker constructor which does this once on instantiation and keeps it as a member variable?

kevinkreiser:
this looks good to me, can anyone give it a quick test? @kdiluca @dnesbitt61 @dgearhart

kevinkreiser:
@yes, the service tests have faked configs and get validated by the worker when its instantiated. so we'll need to patch that into the faked config in the test

ericrwolfe:
Just updated the test. Lmk if there's a cleaner way of adding the array.

Side note: should `meili.customizable` be optional? It's in the default conf/valhalla.json.

kevinkreiser:
doesnt make a difference to me. required is fine i think. @dnesbitt61 was testing this pr with some test requests and i think will merge this afterwards

PR #503, opened 2016-12-07
Add simple A* test

zerebubuth:
Tests that a route with start and end points on the same segment will take the shortest path along that segment rather than a detour along a longer route. This test was written trying to track down a bug where that wasn't the case.

@kevinkreiser could you review, please?


kevinkreiser:
i love you for doing this! super mega :ship:

PR #504, opened 2016-12-07
Increase edge length tolerance and set a minimum for short edges.

dnesbitt61:
Tolerance was too tight on last commit, it broke some cases.


dgearhart:
:ship:

PR #510, opened 2016-12-08
Added Filter Attributes & Shape Match Algorithm Choices

kdiluca:
- Added capability to include/exclude the edge attributes
- Added a string shape_match to choose which shape algorithm to run; either "edge_walking", "map_snap" (only runs map-matching) or "walk_or_snap" (runs edge_walking first and falls back to map_matching).
- Serializes out the attributes from the trip_path proto based on the include/exclude parameters with default

Example request:
 http://localhost:8002/trace_attributes?json={"shape":[{"lat":39.983841,"lon":-76.735741},{"lat":39.983704,"lon":-76.735298},{"lat":39.983578,"lon":-76.734848},{"lat":39.983551,"lon":-76.734253},{"lat":39.983555,"lon":-76.734116},{"lat":39.983589,"lon":-76.733315},{"lat":39.983719,"lon":-76.732445},{"lat":39.983818,"lon":-76.731712},{"lat":39.983776,"lon":-76.731506},{"lat":39.983696,"lon":-76.731369}],"costing":"auto","shape_match":"walk_or_snap","filters":{"attributes":["edge.names","edge.id", "edge.weighted_grade","edge.speed"],"action":"only"}}


src/thor/trace_attributes_action.cc
```diff
@@ -27,39 +26,108 @@ namespace {
   const headers_t::value_type JSON_MIME { "Content-type", "application/json;charset=utf-8" };
   const headers_t::value_type JS_MIME { "Content-type", "application/javascript;charset=utf-8" };
 
-  json::MapPtr serialize(odin::TripPath trip_path, const boost::optional<std::string>& id, double scale) {
+
+  json::MapPtr serialize(TripPathController& controller, valhalla::odin::TripPath trip_path, const boost::optional<std::string>& id, double scale) {
```
dgearhart: trip_path can be passed by const ref
const valhalla::odin::TripPath&

src/thor/trace_attributes_action.cc
```diff
@@ -27,39 +26,108 @@ namespace {
...
+        if (edge.has_end_shape_index())
+          edgemap->emplace("end_shape_index", static_cast<int64_t>(edge.end_shape_index()));
+        if (edge.has_begin_shape_index())
+          edgemap->emplace("begin_shape_index", static_cast<int64_t>(edge.begin_shape_index()));
+        if (edge.name_size() > 0) {
+          for (const auto& name : edge.name())
+            names->push_back(name);
+          edgemap->emplace("names", names);
+        }
+        if (controller.node_attribute_enabled()) {
```
dnesbitt61: A couple questions:
1) Other places look for presence of an item in the trip path protocol buffer. Could this be done here?
2) Casts to int64_t and uint64_t - do these have to be 64 bit or can plain int be used? Outside of way_id and edge Ids most fields do not need 64 bits.

src/thor/trace_attributes_action.cc
```diff
@@ -27,39 +26,108 @@ namespace {
...
-      json->emplace("id", *id);
-    return json;
   }
 }
 
 namespace valhalla {
 namespace thor {
 
+void thor_worker_t::filter_attributes(const boost::property_tree::ptree& request, TripPathController& controller) {
+  std::unordered_set<std::string> attributes_include_;
```
dgearhart: Are these needed? Can update controller in just one loop

src/thor/trace_attributes_action.cc
```diff
@@ -27,39 +26,100 @@ namespace {
...
-  json::MapPtr serialize(odin::TripPath trip_path, const boost::optional<std::string>& id, double scale) {
+
+  json::MapPtr serialize(TripPathController& controller, valhalla::odin::TripPath& trip_path, const boost::optional<std::string>& id, double scale) {
+    //lets get some edge attributes
     json::ArrayPtr edges = json::array({});
     for (int i = 1; i < trip_path.node().size(); i++) {
+
       if (trip_path.node(i-1).has_edge()) {
         const auto& edge = trip_path.node(i - 1).edge();
         auto names = json::array({});
```
dgearhart: move down inside if where node is checked

dgearhart:
:shipit:

PR #513, opened 2016-12-09
Improve the length tolerance computation

dnesbitt61:
(minimum was too low for some edges). If expanding from a node fails, do not return, rather pop the last edge added to path_infos and proceed to next edge. This should allow recovery if a wrong path is taken (due to tolerances being met but actually on wrong path).


dgearhart:
:ship:

kevinkreiser:
@noblige give this a try!

PR #517, opened 2016-12-09
Fix remaining distance for trivial paths

zerebubuth:
The remaining distance for trivial paths made up of only a single edge was being set to the distance between the origin and destination. This was causing trivial edges to be assigned effectively twice the distance between the origin and destination, which allowed some alternate paths which were shorter than twice the minimum distance to be selected.

There's also a test case to check that this is working as expected.

@dnesbitt61 could you review, please?


dnesbitt61:
:shipit:

PR #519, opened 2016-12-11
Let meili::FindOriginMatch find the origin

ptpt:
See https://github.com/valhalla/meili/pull/64


src/thor/trace_route_action.cc
```diff
@@ -165,14 +166,11 @@ odin::TripPath thor_worker_t::map_match(const TripPathController& controller) {
...
-      results.begin(), results.end(),
-      [](const meili::MatchResult& result) {return result.HasState();});
-  auto last_result_with_state = std::find_if(
-      results.rbegin(), results.rend(),
-      [](const meili::MatchResult& result) {return result.HasState();});
+  auto first_result_with_state = meili::FindOriginMatch(results.begin(), results.end());
+  auto last_result_with_state = meili::FindDestinationMatch(results.begin(), results.end());
   if ((first_result_with_state != results.end())
       && (last_result_with_state != results.rend())) {
+
```
dnesbitt61: Seems to need to be: && (last_result_with_state != results.end()))  to compile for me?

ptpt: Yes, updated. Sorry about it!

I actually didn't make thor successfully compiled last time. Will give it a try now.

ptpt:
Close because it doesn't fix the issue. https://github.com/valhalla/thor/pull/525 should fix.

PR #520, opened 2016-12-12
compatibility for libprime_server 0.6.0

kevinkreiser:


dgearhart:
:ship: when circle happy

PR #525, opened 2016-12-13
Ensure that origin pathlocations contains the first pathedge

ptpt:
Second trial on https://github.com/valhalla/meili/pull/64 based on @dnesbitt61's hack. Should be more robust.


dnesbitt61:
That looks good to me!  :shipit:

dnesbitt61:
:shipit:

PR #532, opened 2016-12-16
Add interrupt to map-matching

dnesbitt61:
When called through thor service. Depends on https://github.com/valhalla/meili/pull/66


dgearhart:
:shipit:

PR #535, opened 2016-12-19
Restricted the output for the intersecting edge walkability, cyclability, and driveability.

dgearhart:


kdiluca:
:christmas_tree:

PR #541, opened 2017-01-06
multimodal - Fallback to a pedestrian only route (no transit) 

dnesbitt61:
if destination is not within the max walking distance from a transit stop, but the origin and
dest. locations are within 2km (later have this use the costing options). This allows short routes to succeed as a pedestrian only fallback when no transit is near the origin or dest locations.


src/thor/multimodal.cc
```diff
@@ -120,6 +112,23 @@ std::vector<PathInfo> MultiModalPathAlgorithm::GetBestPath(
...
+  // destination - for now assume pedestrian
+  // TODO - some means of setting destination mode
+  bool disable_transit = false;
+  if (!CanReachDestination(destination, graphreader, TravelMode::kPedestrian, pc)) {
+    // Return if distance exceeds maximum distance set for the starting distance
+    // of a multimodal route (TODO - add methods to costing to support this).
+    if (mindist > 2000) {
+      // Throw an exception so the message is returned in the service
+      throw valhalla_exception_t{400, 440};
+      return { };
```
dgearhart: is the return needed is you are throwing exception?

dgearhart:
:ship:

PR #542, opened 2017-01-13
change from info to debug

kevinkreiser:
all the info about cost and iterations is great debug info but gets quite noisy when you do it a lot. earlier when we discussed logging we mentioned that this info was only for someone debugging things. so i've gone ahead and demoted these messages to debug.

NOTE: @kdiluca @dgearhart i didnt touch the plethora of LOG_INFO found in the various trace_* code. I figure we'll be cleaning that up before release?


dnesbitt61:
:shipit:

kdiluca:
Correct I will be cleaning all of that up along with doing the other issues that I added for logging.

PR #544, opened 2017-01-19
Configurable source to target algorithm

TeXitoi:
Hi!

Here is a proposition for configuring the algorithm of sources_to_targets. A PR in conf and tools is also needed.

I kept it simple, sufficient for our needs. I let you say if we should do something else, if we don't do the valhalla way or any other suggestions.


dnesbitt61:
I think we might want to override on a case by case basis, rather than (or in addition to) a config variable. For example, I think we always want to use CostMatrix for auto but probably want to use TimeDistanceMatrix for pedestrian and bicycle. It might also depend on the number of locations and the distance between them (for small number of locations more widely distributed it may make sense to use CostMatrix).

src/thor/matrix_action.cc
```diff
@@ -125,9 +126,21 @@ namespace valhalla {
 
       json::MapPtr json;
       //do the real work
-      thor::CostMatrix costmatrix;
+      std::vector<TimeDistance> time_distances;
+      switch (source_to_target_algorithm) {
```
kevinkreiser: so then may we just change this to be an `if` `else` to match what @dnesbitt61  was saying instead of to get it from config?

kevinkreiser: you know like:

```
if bike || ped && locations are close
  ...
else
  ...
```

dnesbitt61: I would add a third case so you have COST_MATRIX, TIME_DISTANCE_MATRIX, and (default) SELECT_OPTIMAL. Thus the conf could override and use the specified one or it could rely on code to select the best (based on costing type, other considerations). This would satisfy both needs I think.

TeXitoi:
So I added a SELECT_OPTIMAL. I use timedistance if More than 100 locations (costmatrix is not memory friendly when there is a lot of requests).

Then, I select according to the mode. I don't have any idea of "locations are close", thus I didn't use this criterion for the moment. If you have any idea of such a definition, I'll happily implement it.

src/thor/matrix_action.cc
```diff
@@ -125,9 +126,40 @@ namespace valhalla {
...
+      };
+      auto timedistancematrix = [&]() {
+        thor::TimeDistanceMatrix matrix;
+        return matrix.SourceToTarget(correlated_s, correlated_t, reader, mode_costing, mode);
+      };
+      switch (source_to_target_algorithm) {
+      case SELECT_OPTIMAL:
+        if (correlated_s.size() + correlated_t.size() > 100) {
+          time_distances = timedistancematrix();
+        } else {
```
TeXitoi: Mmm, as you've merged, should I make this change?

dnesbitt61:
Hi - I will make a quick change to comment this out (other than the size comparison) with a TODO comment about needing to run some performance tests to characterize which algorithm is optimal for different cases. Will that be OK? We are preparing a release candidate for later this week and this looked nearly ready.

TeXitoi:
That's OK for us, we will explicitly use timedistancematrix on our instance, at least for the moment.

PR #545, opened 2017-01-19
Bug fix in CanReachDestination

dnesbitt61:
Fix bug where lambda in CanReachDestination was using class member edgelabels_ rather than the locally defined edgelabels. It is unlikely this was ever called but best to be correct and safe.


dgearhart:
:ship:

